PR #76, opened 2024-04-06
Add coroutine support to btl::futures

samienne:


samienne:
Closing as stale. This has sat unattended long enough that it would need rewriting against the current `btl::future` code rather than rebasing — `merge.h` and `futurecontrol.h` have both changed since (#104, #95). Worth reopening as a fresh PR if coroutine support is wanted.

PR #81, opened 2025-04-19
Add native file for MSVC

samienne:
Create a wrapper to capture vsvarsall environment in the build folder.


samienne:
Closing as obsolete. CI now drives the builds through `lw` (loomworks) using the configuration sets in `loomworks.json` (#97), and `lw` establishes the compiler environment itself — so a native file for MSVC is no longer the mechanism. Reopen if the wrapper turns out to cover a case `lw` does not.

PR #85, opened 2026-07-12
CI: add clang-cl to the Windows build matrix

samienne:
Adds a second Windows CI leg that builds with **clang-cl**, alongside the existing MSVC build, plus the source fix required to make that build link.

## Commits
- **Fix template symbol export for clang-cl on Windows** — the Windows visibility macros left `BTL_EXPORT_TEMPLATE`/`BTL_IMPORT_TEMPLATE` empty, so the extern-template export scheme was a no-op. MSVC tolerated it by re-emitting inline members in each consumer; clang-cl honors `extern template` strictly and expects the symbols from the DLL, which never exported them → unresolved `Animated<T>`, `Widget<>`, `Element<>`, modifier templates. Marks the extern-template decls dllimport/dllexport and puts the export attribute on the explicit instantiation definitions.
- **CI: build Windows with both MSVC and clang-cl** — converts `build-win` into a `{msvc, clang-cl}` matrix.

## Why together
The clang-cl CI leg only links because of the export fix, so they belong in one self-contained PR (independent of the shape-utilities work in #84).

## Testing
Verified locally on Windows with clang-cl 18 and MSVC 19.43 (both build + link clean). This PR's own `build-win (clang-cl)` leg is the CI validation.

### DIFF
--- a/src/btl/include/btl/visibility.h
+++ b/src/btl/include/btl/visibility.h
@@ -27,8 +27,8 @@
 #elif defined(_WIN32)
     #define BTL_EXPORT __declspec(dllexport)
     #define BTL_IMPORT __declspec(dllimport)
-    #define BTL_EXPORT_TEMPLATE
-    #define BTL_IMPORT_TEMPLATE
+    #define BTL_EXPORT_TEMPLATE __declspec(dllexport)
+    #define BTL_IMPORT_TEMPLATE __declspec(dllimport)
     #define BTL_EXPORT_CLASS_TEMPLATE
 
     #define BTL_FORCE_INLINE inline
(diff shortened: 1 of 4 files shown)


src/btl/include/btl/visibility.h
```diff
@@ -27,8 +27,13 @@
 #elif defined(_WIN32)
     #define BTL_EXPORT __declspec(dllexport)
     #define BTL_IMPORT __declspec(dllimport)
-    #define BTL_EXPORT_TEMPLATE
-    #define BTL_IMPORT_TEMPLATE
+    // Export/import explicit template instantiations via the extern-template
+    // declaration. clang-cl honors extern template strictly and will not
+    // re-emit members locally, so the instantiation must be marked; MSVC
+    // accepts the same annotation. The class template itself stays unmarked
+    // (CLASS_TEMPLATE empty), mirroring the Clang branch above.
```
samienne: I think these comments are quite excessive and unnecessary. We can just remove these.

PR #87, opened 2026-07-13
Rewrite the readme as a user-facing readme.md

samienne:
Replaces the stale `readme.org` with a user-facing `readme.md`.

The old readme used the pre-rename `reactive::` namespace, claimed "Linux only", had broken org-mode blocks, and none of its examples compiled.

The new one is purely user-facing:
- Intro (the only place other frameworks are mentioned), build instructions for Linux + Windows, platform table.
- Concept sections — signals, streams, widgets & layout, modifiers, shapes, animation — each with a short example **verified against the current API**.
- A complete counter example.
- No internal architecture details.

Best reviewed rendered. Happy to adjust tone/coverage.


readme.md
```diff
@@ -0,0 +1,308 @@
+# Reactive
+
+[![build](https://github.com/samienne/reactive/actions/workflows/build-workflow.yaml/badge.svg)](https://github.com/samienne/reactive/actions/workflows/build-workflow.yaml)
+
+Reactive is a C++17 graphical user interface toolkit built on **functional
```
samienne, line 5: Actually Reactive was the old name (the repo is still named reactive also) as I started this library so long ago that React was not a thing (or I didn't know about). So I decided to rename this as Bq (Becquerell) to continue the reactive/nuclear theme, although the name is a bit difficult and there is a bit confusion on gui and qui in the name. Also Bq is the library that handles only the signals and the reactive part and bqui is the ui library built on top of it. I might rename this later on.

readme.md
```diff
@@ -0,0 +1,308 @@
...
+
+### Requirements
+
+- **Meson** and **Ninja**
+- A C++17 compiler — GCC 10+, Clang 11+, MSVC 2022, or clang-cl
+- On Linux, the OpenGL development headers (e.g. `libgl1-mesa-dev`)
+
+Third-party dependencies (Eigen, FreeType, and a few others) are downloaded
+automatically by Meson the first time you configure, so an internet connection
+is needed for the initial setup.
```
samienne, line 63: I don't know if we should mention that Meson is capable of using platform provided versions of these as well. This is pretty standard Meson stuff so it might be obvious to those who actually use it and for others it is convenient that it is downloaded automatically if needed. Either way I'm fine.

readme.md
```diff
@@ -0,0 +1,308 @@
...
+
+You derive new signals from existing ones with `map`. The derived signal updates
+automatically whenever its source does:
+
+```cpp
+auto text = count.signal.map([](int n)
+    {
+        return std::to_string(n);
+    });
+// text is a Signal<std::string> that always mirrors count
```
samienne: There's a small distinction here that AnySignal<std::string> is the type erased counterpart of Signal<Storage, std::string> where the Storage is some internal data type holding everything needed for the signal to exist. This is purely internal thing so no need to explain it here but I think we should use the AnySignal regardless. I admit that the naming might be a bit confusing at first but this library uses the Any prefix to denote type erasure.

readme.md
```diff
@@ -0,0 +1,308 @@
...
+
+```cpp
+#include <bqui/withanimation.h>
+#include <avg/curve/curves.h>
+
+{
+    auto anim = withAnimation(0.3f, avg::curve::easeInOutCubic);
+    open.handle.set(true);      // this change animates over 0.3 seconds
+}
+// changes made outside the scope are applied immediately
```
samienne, line 256: There are other variants of withAnimation and the one used in this example uses the guard style where the animation is applied when the destructor of the guard is run. There's also a lambda variant. It might be useful to mention those or at least explain that there's the guard that triggers the animation when it runs out of the scope. Although, no need to be overly thorough.

PR #88, opened 2026-07-13
Add agent-facing documentation scaffolding

samienne:
Bootstraps the project's documentation system, so agents (and contributors) can understand the system without re-reading the code each time.

## Structure
- **Human docs — `readme.md`:** the project readme plus a `src/<lib>/readme.md` per library (concepts + usage). Co-located so they travel if a library becomes its own repository.
- **Agent docs:** `AGENTS.md` at the root — the canonical, tool-neutral hub; `CLAUDE.md` is a one-line import so Claude Code auto-loads it — pointing into `docs/` (architecture overview, conventions, style-guide, decisions, reviewing) and a per-library `src/<lib>/AGENTS.md`.
- **API reference:** Doxygen in the headers (convention established; backfilled incrementally). No signatures are transcribed into Markdown.

## Conventions it establishes
- **Knowledge-base shape:** index → small single-topic files, a soft size cap, one home per fact, docs point at code (never the reverse).
- **Comments** describe the code as it is; history/context goes in a topic file, not a comment.
- **Pre-merge review:** clean-context style + correctness reviews (inputs include `reviewing.md`), a docs-freshness check, **human sign-off required and agents never self-merge**, plus a learning loop that distils human-review lessons into the rubric or `reviewing.md`.

## Dogfooded
Every doc is signature-free and carries a "verified against" stamp. The PR was taken through the review convention it defines — three clean-context reviews, each of which caught a real issue (a wrong `AnySignal` claim, an imprecise type, and a lane-violation in `reviewing.md` itself), all resolved.


docs/architecture.md
```diff
@@ -0,0 +1,165 @@
...
+| `bqui` | The UI toolkit built on all of the above. |
+
+`bq` being independent of the UI is intentional — it is a general reactive
+dataflow library that the UI happens to use.
+
+## The reactive core (`bq`)
+
+### Signals — values over time
+
+A **signal** describes a value that varies over time; a `Signal<int>` is a
```
samienne: Here's again the AnySignal type erasure. This is a bit cumbersome as the user normally works with the type erased types but the Any prefix is needed still. We might visit this at some point but for now I think we should use AnySignal<int>. Actually I realized that this is explained later in this chapter. But let's check this as well.

PR #89, opened 2026-07-14
Widget introspection: names, roles, capabilities, geometry

samienne:
A curated, introspectable description of the widget tree, carried on a widget's realised Instance.

## Introspection node
- Immutable, structurally shared, recursive. Fields: name, role, `Capability` set, `obb`, author data bag, children.
- `Capability`: Clickable, Editable, Focusable, Scrollable. `toString(Capability)` names one.
- `IntrospectionChild` = `shared_ptr<Introspection const>`.
- Curated surface, not a reflection of internals: opaque widgets emit no children; containers aggregate; an affordance whose geometry diverges from its parent becomes a child node.

## API
- Set via modifiers: `setName`, `setRole`, `setData`, `addCapability`, `setWidgetIntrospection`.
- `obb` is in the node's local space; `resolveIntrospection` flattens to absolute window space at the consumer boundary.

## Notes
- Read off the widget's realised Instance; no widget internals are exposed.
- Groundwork for headless / agent inspection (built on by #96).

bq untouched.

### DIFF
--- a/src/bqui/include/bqui/modifier/setwidgetintrospection.h
+++ b/src/bqui/include/bqui/modifier/setwidgetintrospection.h
@@ -1,0 +1,28 @@
+#pragma once
+
+#include "widgetmodifier.h"
+
+#include "bqui/widget/introspection.h"
+#include "bqui/bquivisibility.h"
+
+#include <bq/signal/signal.h>
+
+#include <string>
+
+namespace bqui::modifier
+{
+    // Replace the widget's introspection with the given (reactive) value. This
+    // is the primitive; the more specific modifiers below are built on it.
+    BQUI_EXPORT AnyWidgetModifier setWidgetIntrospection(
+            bq::signal::AnySignal<widget::Introspection> introspection);
+
+    BQUI_EXPORT AnyWidgetModifier setWidgetIntrospection(
+            widget::Introspection introspection);
+
+    // Give the widget a name, carried along the build. Transforms the current
+    // introspection rather than replacing it, so an existing role survives.
+    BQUI_EXPORT AnyWidgetModifier setName(std::string name);
+
+    // Set the widget's role (its "what kind" tag), e.g. "ScrollBar".
+    BQUI_EXPORT AnyWidgetModifier setRole(std::string role);
+} // namespace bqui::modifier
--- a/src/bqui/include/bqui/modifier/setwidgetintrospection.h
+++ b/src/bqui/include/bqui/modifier/setwidgetintrospection.h
@@ -3,6 +3,7 @@
 #include "widgetmodifier.h"
 
 #include "bqui/widget/introspection.h"
+#include "bqui/widget/datavalue.h"
 #include "bqui/bquivisibility.h"
 
 #include <bq/signal/signal.h>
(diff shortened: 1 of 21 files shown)


src/bqui/include/bqui/modifier/setwidgetintrospection.h
```diff
@@ -0,0 +1,66 @@
...
+
+#include <string>
+#include <utility>
+
+namespace bqui::modifier
+{
+    // Replace the widget's introspection with the given (reactive) value. This
+    // is the primitive; the more specific modifiers below are built on it.
+    inline auto setWidgetIntrospection(
+            bq::signal::AnySignal<widget::Introspection> introspection)
```
samienne: We could do here that we return AnyWidgetModifier and just put this into a .cpp file to make the compilation faster or we could do a full template template <typename T> auto setWidgetIntroSpection(Signal<T, Introspection> introspection).

src/bqui/src/widget/layout.cpp
```diff
@@ -87,11 +92,29 @@ widget::AnyWidget layout(SizeHintMap sizeHintMap,
                         )(params));
         }
 
+        // Aggregate the children's introspection into this layout's node.
+        // handleGravity preserves each child's introspection, so we can read
+        // it straight off the built builders.
```
samienne: I think this is quite obvious from the code already so I don't think we need the comment here.

src/testapp1/main.cpp
```diff
@@ -69,6 +75,31 @@ std::vector<std::pair<std::string, avg::Curve>> curves = {
...
+
+        for (auto const& child : node.children)
+            printIntrospection(child, depth + 1);
+    }
+
+    void printWidgetHierarchy(widget::AnyWidget widget)
+    {
+        auto introspection = std::move(widget)(BuildParams{}).getIntrospection();
+        auto data = bq::signal::makeSignalContext(std::move(introspection))
+            .evaluate();
```
samienne: I know this is now just test code but I would like to warn about this as I have actually ran into this same urge to just whip out a quick signal context to evaluate some signal. The main danger comes from the more dynamic elements (especially the streams) and when a signal context gets built. The signal context is completely parallel and the signal states don't actually share with each other (there's a way to force this for a specific signals but it has to be very deliberate). So for example, if a signal has stream::iterate in its evaluation chain and we then send some events into the stream the signal will change its value, however, it is important to note that when we get a new signal context after these events the context doesn't evaluate into the same state as the events have not been seen by the new context. So if we use a stream to add new widgets into the tree those might not end up being visible. Just something to make a note about for later. We will most likely get rid of this function once we have the proper setup.

src/bqui/include/bqui/modifier/setwidgetintrospection.h
```diff
@@ -0,0 +1,53 @@
...
+
+namespace bqui::modifier
+{
+    /// Replace the widget's introspection with the given (reactive) value,
+    /// keeping the instance's realised obb. The full escape hatch; the more
+    /// specific modifiers below build on it.
+    BQUI_EXPORT AnyWidgetModifier setWidgetIntrospection(
+            bq::signal::AnySignal<widget::Introspection> introspection);
+
+    /// \overload
```
samienne: Use c-style comments /** and @overload.

src/bqui/include/bqui/widget/datavalue.h
```diff
@@ -0,0 +1,54 @@
...
+            double,
+            bool,
+            DataMap,
+            DataArray
+            >;
+
+        Variant value = std::string{};
+
+        DataValue() = default;
+        DataValue(std::string value) : value(std::move(value)) {}
```
samienne: Let's put these in a .cpp file.

src/ase/include/ase/dummyplatform.h
```diff
@@ -4,18 +4,59 @@
...
+#include <chrono>
+#include <memory>
+#include <vector>
+
 namespace ase
 {
+    class DummyWindow;
+    class Platform;
+
+    /// A headless platform backend. It creates no OS window and drives a
```
samienne: Always use the c-style /** comments with doxygen. I would also prefer to have the @brief at the top.

src/ase/include/ase/dummywindow.h
```diff
@@ -37,10 +38,16 @@ namespace ase
         void setHoverCallback(std::function<void(HoverEvent const&)> cb) override;
         void setTextCallback(std::function<void(TextEvent const&)> cb) override;
 
+        // Drive the stored frame callback for one frame. The headless platform
+        // calls this to advance a window; input injection reuses the same
+        // GenericWindow seam the real backends use.
+        bool needsRedraw() const;
+        std::optional<std::chrono::microseconds> frame(Frame const& frame);
+        GenericWindow& getGenericWindow();
```
samienne: I dislike the idea of exposing GenericWindow as it is mainly meant for implementations to simplify bookkeeping for the windows. Also I think we would like to support injecting events to all types of windows including the gl ones. I wonder if we need to put the inject methods into the WindowImpl base class.

src/ase/include/ase/dummyplatform.h
```diff
@@ -4,18 +4,59 @@
...
+
+        /// Cap the number of frames run() will pump. Zero means unbounded (only
+        /// the app callback returning false stops it). Default is bounded so a
+        /// headless run cannot spin forever.
+        void setMaxFrames(uint64_t maxFrames);
+
+        /// Advance exactly one frame: run the app callback and each window
+        /// frame with the fixed dt. Returns whether the app callback asked to
+        /// keep running. Lets a caller pump frames on demand.
+        bool step(std::function<bool(Frame const&)> const& frameCallback);
```
samienne: I would actually want all of the platforms behaving identically so I would want all of these methods to be in the WindowImpl. I would like to be able to drive even the gl window from the agent interface. We might have to think about this throughly as some platforms (e.g. ios) take control of the main loop so we might need some kind of smarter interface here.

src/ase/include/ase/dummyplatform.h
```diff
@@ -4,18 +4,81 @@
...
+         * @brief Advance exactly one frame at the supplied time: run the app
+         * callback, then each live window's frame callback.
+         *
+         * Clock-agnostic — the caller supplies the frame — so run() pumps fixed
+         * deterministic frames and an external driver can supply its own.
+         *
+         * @return Whether the app callback asked to keep running.
+         */
+        bool step(Frame const& frame,
+                std::function<bool(Frame const&)> const& frameCallback) override;
```
samienne: Do we really need the frame callback here? I was thinking that the step would just "inject" time into the system. The callback is used in the run() loop.

src/ase/src/dummy/dummyplatform.cpp
```diff
@@ -29,11 +33,52 @@ RenderContext DummyPlatform::makeRenderContext()
...
+    maxFrames_ = maxFrames;
+}
+
+bool DummyPlatform::step(Frame const& frame,
+        std::function<bool(Frame const&)> const& frameCallback)
+{
+    handleEvents();
+
+    if (!frameCallback(frame))
+        return false;
```
samienne: As I assumed, this could easily be handled outside step and inside run to keep the api surface minimal.

src/bqui/src/app.cpp
```diff
@@ -472,9 +529,50 @@ App App::windows(std::initializer_list<Window> windows) &&
...
+App App::agentic(bool agentic) &&
+{
+    d()->agentOverride_ = agentic;
+    return std::move(*this);
+}
+
 int App::run(bq::signal::AnySignal<bool> runningSignal) &&
 {
-    ase::Platform platform = ase::makeDefaultPlatform();
+    // Platform: explicit override, else headless env, else the OS default.
```
samienne: No need for obvious comments.

src/testapp1/main.cpp
```diff
@@ -69,6 +75,65 @@ std::vector<std::pair<std::string, avg::Curve>> curves = {
     { "easeInOutBounce", avg::curve::easeInOutBounce },
 };
 
+namespace
+{
+    char const* capabilityName(widget::Capability cap)
```
samienne: Would actually be nice to have this in the same place where widget::Capability is located. Maybe name it to to_string or toString.

PR #92, opened 2026-07-15
Generalize SignalContext to N signals over one shared DataContext

samienne:
## What

Generalizes `bq::signal::SignalContext` from a **single** signal to **N signals sharing one `DataContext`**, so multiple top-level signals update in lockstep and `.share()`/`.cache()` nodes deduplicate across them. This is the reactive-core groundwork for driving several signals (widget instance, introspection, window title) from one live context.

`SignalContext` is now parameterized over signal **types** (`SignalContext<TSignals...>`, each a `Signal<...>` or `AnySignal<...>`), not value packs. All entries share one `DataContext`.

## Design

- **One shared `DataContext`** across all entries — a `.share()` node reachable from more than one entry is evaluated **once per pass**, not once per entry. `.share()` is the cross-signal dedup lever; `.cache()` remains the temporal/sub-expression memoization lever.
- **Parameterized over signal types** (not value packs) so typed and type-erased signals coexist in one context, and it stays distinct from `merge`/`combine` (which fuse into one value pack).
- **Indexed accessors, uniformly**: `evaluate<I>()`, `didChange<I>()`, `didAnyChange()`. `evaluate` is **always indexed** — `evaluate<0>()` even with a single signal (no non-indexed convenience, no arity branching).
- **Results cached by owned value.** Every entry's result is evaluated on construction and refreshed after any update in which that entry changed; `emplace` copies the (possibly reference-holding) evaluate result into the owned `SignalResult`, so the cache never aliases signal data and can't go stale. `evaluate<I>()` returns a const reference into that owned cache — a value obtained from it is a snapshot, not a live view.
- **Cadence deliberately dropped**: every entry updates every frame; per-signal evaluation memoization is the user's explicit choice via `.cache()`.
- **The one hard correctness constraint**: `DataContext::swapFrameData()` runs **exactly once per update pass, after every entry has updated** — never per entry, or one entry would rotate away frame data another still needs.

## Migration

- The non-indexed `evaluate()`/`didChange()` are gone; all single-signal call sites now index explicitly (`evaluate<0>().get<0>()`, `didChange<0>()`).
- `makeSignalContext(sigs...)` deduces `SignalContext<std::decay_t<TSignals>...>` and accepts one or many signals.
- `WindowGlue`'s two contexts become the `AnySignal<...>` one-arg form; keeping them as two separate contexts is intentional (combining them is a separate future PR). No `extern template` instantiations of `SignalContext` exist, so none needed updating.
- Because results are owned snapshots (not live references), the two tests that relied on live-reference behavior were updated: `signal.merge`'s result type drops the `const&`s (`SignalResult<int, int, std::string>`), and `signal.join` reads the post-update value off a freshly re-evaluated snapshot rather than an earlier one.

## Tests

New `src/bq/test/signalcontexttest.cpp`: multi-signal lockstep updates; the shared-subgraph-evaluated-once-per-pass guarantee; single-signal use.

Full suite green in a configured MSVC build: `bq`, `avg`, `bqui`, `btl` — **4/4 OK** (new tests plus the existing `share`/`merge`/`join`/`cache`/`check` suite, migrated to the indexed accessors).

## Docs

- `docs/decisions.md`: records the decision (multi-signal over one shared `DataContext`; signal-type parameterization; dropped cadence; `.share()` vs `.cache()` levers; owned-value result caching; indexed-only `evaluate<I>()`; the `swapFrameData`-once-per-pass constraint).
- `src/bq/AGENTS.md`: notes the multi-signal capability in the evaluation model.

### DIFF
--- a/src/bq/include/bq/signal/signalcontext.h
+++ b/src/bq/include/bq/signal/signalcontext.h
@@ -2,93 +2,133 @@
 
 #include "signal.h"
 #include "datacontext.h"
+#include "signaltraits.h"
+
+#include <tuple>
+#include <array>
+#include <cstddef>
+#include <utility>
 
 namespace bq::signal
 {
-    template <typename... Ts>
+    /** Drives and evaluates a set of signals.
+     *
+     * Signals are pure values: code builds and transforms them (constructing,
+     * mapping, combining) and normally never touches a SignalContext directly.
+     * The context is the one place the signal graph is actually run — it owns
+     * the per-signal state, advances it once per frame with `update()`, and
+     * hands back the current values via `evaluate<I>()` (each signal addressed
+     * by its index). In a framework this evaluation is centralized in a single
+     * long-lived context; bqui drives a window this way.
+     *
+     * Prefer one long-lived context over throwaway ones. The state lives in the
+     * context, so evaluating a signal in a fresh short-lived context starts from
+     * scratch and won't reflect what a live context has accumulated (stream
+     * history in particular is not replayed) — reach for the running context,
+     * not a quick one made to peek at a value.
+     *
+     * Several signals can share one context, so a `.share()` node reachable from
+     * more than one of them is evaluated once per update, not once per signal.
+     */
+    template <typename... TSignals>
     class SignalContext
     {
-    public:
-        using InnerResultType = std::optional<SignalResult<Ts...>>;
+        template <size_t I>
+        using EntryResultT =
+            SignalTypeT<std::tuple_element_t<I, std::tuple<TSignals...>>>;
 
-        SignalContext(AnySignal<Ts...> sig) :
-            sig_(std::move(sig)),
-            data_(sig_.unwrap().initialize(dataContext_, FrameInfo(0, {}))),
-            result_(sig_.unwrap().evaluate(dataContext_, data_))
+    public:
+        SignalContext(TSignals... sigs) :
+            signals_(std::move(sigs)...),
+            data_(initializeData(std::index_sequence_for<TSignals...>{})),
+            results_(evaluateAll(std::index_sequence_for<TSignals...>{}))
         {
             dataContext_.swapFrameData();
-        };
-
-        auto evaluate() const -> decltype(auto)
-        {
-            if constexpr(sizeof...(Ts) == 0)
-                sig_.unwrap().evaluate(dataContext_, data_);
-            if constexpr (sizeof...(Ts) == 1)
-                return result_->template get<0>();
-            else
-                return SignalResult<Ts...>(*result_);
         }
 
         UpdateResult update(FrameInfo const& frame)
         {
-            auto r = sig_.unwrap().update(dataContext_, data_, frame);
-            if (r.didChange)
-                result_ = sig_.unwrap().evaluate(dataContext_, data_);
-            dataContext_.swapFrameData();
-
-            didChange_ = r.didChange;
-
-            return r;
+            return updateImpl(frame, std::index_sequence_for<TSignals...>{});
         }
 
+        template <size_t I>
+        EntryResultT<I> const& evaluate() const
+        {
+            return std::get<I>(results_);
+        }
+
+        template <size_t I>
         bool didChange() const
         {
-            return didChange_;
+            return changed_[I];
         }
 
-        btl::connection observe(std::function<void()> callback)
+        bool didAnyChange() const
         {
-            return sig_.unwrap().observe(dataContext_, data_,
-                    std::move(callback));
-        }
-
-        template <typename... Us, typename = std::enable_if_t<
-            btl::all(std::is_convertible_v<Ts, Us>...)
-            >>
-        operator SignalContext<Us...>() const
-        {
-            return SignalContext<Us...>(sig_);
+            for (bool c : changed_)
+                if (c)
+                    return true;
+            return false;
         }
 
     private:
-        AnySignal<Ts...> sig_;
-        DataContext dataContext_;
-        typename AnySignal<Ts...>::DataType data_;
-        InnerResultType result_;
-        bool didChange_ = false;
+        template <size_t... Is>
+        std::tuple<SignalDataTypeT<TSignals>...> initializeData(
+                std::index_sequence<Is...>)
+        {
+            return std::tuple<SignalDataTypeT<TSignals>...>(
+                    std::get<Is>(signals_).unwrap().initialize(
+                        dataContext_, FrameInfo(0, {}))...);
+        }
+
+        // Evaluate one entry and return its result by owned value (copying any
+        // references the signal returns into the cache's SignalTypeT).
+        template <size_t I>
+        EntryResultT<I> evaluateEntry() const
+        {
+            return std::get<I>(signals_).unwrap().evaluate(
+                    dataContext_, std::get<I>(data_));
+        }
+
+        template <size_t... Is>
+        std::tuple<EntryResultT<Is>...> evaluateAll(std::index_sequence<Is...>)
+        {
+            return std::tuple<EntryResultT<Is>...>(evaluateEntry<Is>()...);
+        }
+
+        template <size_t I>
+        UpdateResult updateEntry(FrameInfo const& frame)
+        {
+            auto r = std::get<I>(signals_).unwrap().update(
+                    dataContext_, std::get<I>(data_), frame);
+            changed_[I] = r.didChange;
+            if (r.didChange)
+                std::get<I>(results_) = evaluateEntry<I>();
+            return r;
+        }
+
+        template <size_t... Is>
+        UpdateResult updateImpl(FrameInfo const& frame, std::index_sequence<Is...>)
+        {
+            UpdateResult result = (UpdateResult{} + ... + updateEntry<Is>(frame));
+            dataContext_.swapFrameData();
+            return result;
+        }
+
+        // evaluate() must be usable on a const SignalContext, but the signal
+        // interface takes DataContext by non-const reference for shared/cached
+        // lookups. Evaluation does not mutate signal state, so the context is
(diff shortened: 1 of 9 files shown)


src/bq/include/bq/signal/signalcontext.h
```diff
@@ -2,93 +2,222 @@
...
         }
 
-        UpdateResult update(FrameInfo const& frame)
+        // The actual SignalResult a signal produces when evaluated, as opposed
+        // to SignalTypeT which reports the declared value pack.
+        template <typename TSignal>
+        using EvaluatedResultT = std::decay_t<decltype(
+                std::declval<TSignal const&>().unwrap().evaluate(
+                    std::declval<DataContext&>(),
+                    std::declval<SignalDataTypeT<TSignal> const&>()))>;
```
samienne: There should be already SignalValueType and SignalType in signaltraits.h . Can we reuse those?

src/bq/include/bq/signal/signalcontext.h
```diff
@@ -2,93 +2,222 @@
...
+        // reference that survives the context being moved (e.g. reallocated in a
+        // vector); the evaluated reference would otherwise point into the
+        // moved-from data. A multi-value result keeps its element references,
+        // matching the historical evaluate() return type.
+        template <typename TSignal>
+        using SingleCacheResultT = std::conditional_t<
+            SignalResultArity<EvaluatedResultT<TSignal>>::value == 1,
+            DecaySignalResultT<EvaluatedResultT<TSignal>>,
+            EvaluatedResultT<TSignal>
+            >;
```
samienne: This seems a bit excessive. Let's do the same behavior regardless of the arity. If there are too many places to change then let's just do the same behavior for multisignal case by storing the results in the context.

src/bq/include/bq/signal/signalcontext.h
```diff
@@ -2,93 +2,222 @@
...
+        struct SingleResultCache<TSignal>
+        {
+            using type = std::optional<SingleCacheResultT<TSignal>>;
+        };
+
+        template <typename... TSignals>
+        using SingleResultCacheT = typename SingleResultCache<TSignals...>::type;
+    } // namespace detail
+
+    /// A set of top-level signals sharing a single DataContext.
```
samienne: Use doxygen documentation with /** instead of ///

src/bq/include/bq/signal/signalcontext.h
samienne: Removed the arity checks completely — every entry is now cached uniformly by owned value, and evaluate<I>() is always indexed (no non-indexed convenience).

src/bq/include/bq/signal/signalcontext.h
samienne: Reused SignalType: now that results are stored by owned value, SignalTypeT is exactly the cached type, so the custom result-type helpers are gone.

src/bq/include/bq/signal/signalcontext.h
samienne: Switched to /** */ doxygen comments.

src/bq/include/bq/signal/signalcontext.h
```diff
@@ -2,93 +2,128 @@
...
     public:
-        using InnerResultType = std::optional<SignalResult<Ts...>>;
-
-        SignalContext(AnySignal<Ts...> sig) :
-            sig_(std::move(sig)),
-            data_(sig_.unwrap().initialize(dataContext_, FrameInfo(0, {}))),
-            result_(sig_.unwrap().evaluate(dataContext_, data_))
+        SignalContext(TSignals... sigs) :
+            signals_(std::move(sigs)...),
+            data_(initializeData(std::index_sequence_for<TSignals...>{}))
```
samienne: I actually now dislike that we have changed the result_ into optional, thus it can have "no value" state. This makes handling the value a bit more annoying as we are required to check if result_ is valid or not. And e.g. evaluate() signature doesn't really allow us to handle valueless result_. Of course now we have made everything so that we cannot really have valueless result_ but that makes it a bit frustrating because I would like to signal this with the types. The old version immediately evaluated the result_ to have the value which prevented this from happening.

src/bq/include/bq/signal/signalcontext.h
```diff
@@ -2,93 +2,128 @@
...
+            std::get<I>(results_).emplace(std::get<I>(signals_).unwrap().evaluate(
+                    dataContext_, std::get<I>(data_)));
         }
 
-        template <typename... Us, typename = std::enable_if_t<
-            btl::all(std::is_convertible_v<Ts, Us>...)
-            >>
-        operator SignalContext<Us...>() const
+        template <size_t... Is>
+        void evaluateAll(std::index_sequence<Is...>)
```
samienne: I sort of dislike the inplace editing here and I would prefer every method to return the result instead of modifying things in place. This would probably help with removing the optional from the result_ variable as well. Modifying in place is fine if the other option is too messy or cumbersome.

src/bq/include/bq/signal/signalcontext.h
```diff
@@ -2,93 +2,128 @@
...
-        bool didChange_ = false;
-    };
+        template <size_t I>
+        UpdateResult updateEntry(FrameInfo const& frame)
+        {
+            auto r = std::get<I>(signals_).unwrap().update(
+                    dataContext_, std::get<I>(data_), frame);
+            changed_[I] = r.didChange;
+            if (r.didChange)
+                evaluateEntry<I>();
```
samienne: For example here, it is not immediately clear that we are storing the value somewhere.

src/bq/include/bq/signal/signalcontext.h
samienne: Dropped the optional: results_ is now initialized in the constructor's init list (every entry evaluated up front), so a cached result is always present and evaluate() returns it directly — the type no longer admits a valueless state.

src/bq/include/bq/signal/signalcontext.h
samienne: evaluateEntry<I>() now returns the value instead of mutating in place, and the ctor builds results_ from it (which is what let us drop the optional). The only in-place bit left is the refresh assignment on a changing update; the signal's own update() is inherently in-place, so a fully-returning update wasn't feasible.

src/bq/include/bq/signal/signalcontext.h
samienne: Now an explicit results_[I] = evaluateEntry<I>() assignment at the call site, so the store is visible.

src/bq/include/bq/signal/signalcontext.h
```diff
@@ -2,93 +2,133 @@
...
 {
-    template <typename... Ts>
+    /** A set of top-level signals sharing a single DataContext.
+     *
+     * SignalContext is parameterized over signal *types* (each a `Signal<...>`
+     * or `AnySignal<...>`), not value packs. All entries share one DataContext,
+     * so `.share()`/`.cache()` nodes referenced from more than one entry are
+     * evaluated once per pass across the whole set. Every entry is updated on
+     * every `update()`; temporal memoization is the caller's choice via
+     * `.cache()`.
```
samienne: This paragraph feels like it is explaining everything in a historical context. For a new reader I don't think they are thinking "why this doesn't use value packs" because they don't really know the history. The documentation should explain SignalContext more in terms of what it is used for and not the internal details how it used to work and how it works now. The main take away is that SignalContext enables evaluating the signals safely. It can have some inter context remarks that the user shouldn't do a quick context to evaluate things but rely on long lived contexts. Also I would highlight that the context is only to evaluating and driving the signals and normally code mainly operates directly on the signals by constructing, mapping, etc... with them. The evaluation then happens in some centralized place that is usually central to any framework (similar what bqui does).

src/bq/include/bq/signal/signalcontext.h
```diff
@@ -2,93 +2,133 @@
...
+     * update in which that entry changed. `evaluate<I>()` returns entry I's
+     * cached SignalResult; it is always indexed, even with a single signal.
+     */
+    template <typename... TSignals>
     class SignalContext
     {
-    public:
-        using InnerResultType = std::optional<SignalResult<Ts...>>;
+        template <size_t I>
+        using EntryResult =
```
samienne: I would name this as EntryResultT as it is a type and doesn't need the ::type to access it.

src/bq/include/bq/signal/signalcontext.h
```diff
@@ -2,93 +2,133 @@
...
-    template <typename T>
-    struct ResultToContext
-    {
-    };
+        template <size_t... Is>
+        UpdateResult updateImpl(FrameInfo const& frame, std::index_sequence<Is...>)
+        {
+            // Update (and refresh) every entry first, then rotate frame data
+            // exactly once. Rotating per entry would discard frame data a later
+            // entry still needs from the same pass.
```
samienne: These comments are unnecessary as it is obvious from the code.

src/bq/include/bq/signal/signalcontext.h
samienne: Rewrote the class doc around usage: what SignalContext is for (the central place that drives/evaluates signals, with framework code normally operating on the signals directly and bqui as the example), plus the guidance to rely on a long-lived context rather than a throwaway one. Dropped the value-packs/history framing.

src/bq/include/bq/signal/signalcontext.h
samienne: Renamed to EntryResultT.

src/bq/include/bq/signal/signalcontext.h
samienne: Removed those (the ctor and updateImpl swap comments) — the once-per-pass structure reads from the code.

PR #94, opened 2026-07-17
WIP: reproduce + fix whenAll ARM64 cancel race (do not merge)

samienne:
**WIP / debugging-in-progress — do NOT merge.** Attempt to reproduce and fix the rare, ARM64-only `whenAll` cancel-on-failure race on macOS CI.

## TL;DR (honest status)

- A **fix** is applied to `src/btl/include/btl/future/whenall.h` and CI is **green** on the Apple-Silicon runner: run https://github.com/samienne/reactive/actions/runs/29571044242 (`whenAllCancelOnFail` passes 1000 iterations).
- I could **not** produce a *reliably RED, fix-discriminating* reproduction on CI. The one red signal I got (leaks at a short cancellation margin) is a coarse timing artifact that fails **equally for the racy and the fixed code**, so it does not isolate the diagnosed defect. ThreadSanitizer does **not** flag this bug. Details below — please weigh this when reviewing whether the fix is warranted.

## Diagnosis (as handed to me, and what I found)

`whenAll(f1, f2)` must cancel the remaining siblings the instant any sibling fails, so their `.then` continuations never run. Cancellation = `values_.reset()` in `whenall.h`, handed between `init()` (calling thread) and `reportFailure()` (pool thread) via a **store-buffering handshake** on two atomics (`failed_`, `initialized_`), each side storing its own flag then loading the other's with `release`/`acquire` (not `seq_cst`). In theory both loads can miss on ARM64 → neither resets `values_` → the failed future's sibling survives and its `.then` runs.

`src/btl/include/btl/future/merge.h` (`MergedFuture`) has the **identical** handshake — noted for a follow-up, not fixed here.

## The fix

`whenall.h`: replace the two-flag handshake with a single owning happens-before chain — a per-`WhenAll` `SpinLock` guarding every access to `values_` (the `initialized_`/`values_` decision in `init()`, the reset in `reportFailure()`, the move-out in `reportValueReady()`). Exactly one site ever drops `values_`; no load can miss. All existing `btl` async tests still pass locally (clang-cl), including the original `whenAllCancelOnFail` and `mergeFail`.

Also fixed a **pre-existing** bit-rotted TSan annotation in `futurecontrol.h` (`value_.getReferenceForTsan()` — `value_` is a `std::variant` with no such member; only compiled under a TSan build, so it went unnoticed). Legitimate; can stay.

## What the CI iterations actually showed

The macOS job was temporarily reduced to build/run only a small dedicated `asyncstresstest` target (btl is header-only, so ninja compiles just that TU).

| commit | code | sibling sleep / iters | ARM64 result |
|---|---|---|---|
| repro | racy | 40ms / 3000 | **leaked 18** (red) — run 29569112794 |
| fix | fixed | 40ms / 3000 | **leaked 14** (red) — run 29569607152 |
| diag | fixed | 150ms / 400 | 0 (pass) — run 29570372847 |
| diag | racy | 150ms / 400 | 0 (pass) — run 29570560568 |
| diag | racy + TSan detector | (TSan) | 0 races, pass — run 29570804950 |
| final | fixed | 150ms / 1000 | **0 (pass, green)** — run 29571044242 |

Interpretation:
- At a **short** margin (40ms), *both* racy and fixed leak ~equally. That is a **reset-vs-sleep-completion timing race in the harness** under CI load (the sibling's continuation occasionally locks its control before the reset destroys it), independent of the flag handshake — so it does not validate the fix.
- At a **wide** margin (150ms), *both* racy and fixed pass over hundreds of iterations, because the store-buffering **double-miss** is rarer than that. A double-miss also produces **no conflicting memory access**, so **TSan reports nothing** (confirmed on CI).
- Net: I have a clean **green** with the fix, but no clean **red** that the fix flips. The diagnosed race is plausibly real (matches the "intermittent" nature of the original test) but is too rare to catch functionally within a practical CI time budget, and is invisible to TSan.

## TODO before merge

- **Restore `build-linux` and `build-win`** (remove the `if: false` TEMP guards) and restore `ci-success` `needs:` to `[build-linux, build-macos, build-win]`.
- Restore the `build-macos` job's normal full build + `ninja test` (currently reduced to the async-stress target: a debug run + a TSan compile-smoke).
- Decide whether to **keep `asyncstresstest`** (as a bounded regression test) or revert it. As-is it is a weak guard — it passes for both the racy and fixed code at the margin used.
- Decide whether the fix is worth keeping given it could not be tied to a reproduced failure; if kept, apply the same fix to `merge.h`.
- Keep the `futurecontrol.h` `getReferenceForTsan` fix (independently correct).

All TEMP changes carry `# TEMP: ... restore before merge` / `TODO before merge` markers. The intermediate `DIAG`/`TEMP(repro)` commits document the red/green experiments and can be squashed away.

### DIFF
--- a/.github/workflows/build-workflow.yaml
+++ b/.github/workflows/build-workflow.yaml
@@ -71,23 +71,30 @@ jobs:
 
         # TEMP: reduced for ARM64 race debugging — restore before merge.
         # Builds and runs ONLY the small asyncstresstest target (btl is
-        # header-only, so ninja compiles just that TU, not ase/avg/bqui) under
-        # ThreadSanitizer, which flags the whenall.h store-buffering race
-        # deterministically on the Apple-Silicon runner.
-        - name: Build (TSan, async stress only)
+        # header-only, so ninja compiles just that TU, not ase/avg/bqui). A
+        # separate ThreadSanitizer *compile* of the same target keeps the TSan
+        # annotation path building (it had bit-rotted). NOTE: TSan does not flag
+        # this bug — a store-buffering double-miss produces no conflicting
+        # access — so the functional stress test is the detector.
+        - name: Build (async stress only)
           run: |
             sudo rm -Rf /Library/Developer/CommandLineTools
             sudo xcode-select -s /Applications/Xcode.app
+            mkdir -p build/debug
+            cd build/debug
+            python3 ../meson/meson.py setup --buildtype=debug --wrap-mode=forcefallback ../..
+            ninja src/btl/asyncstresstest
+
+        - name: Build (TSan compile smoke test)
+          run: |
             mkdir -p build/tsan
             cd build/tsan
             python3 ../meson/meson.py setup --buildtype=debug --wrap-mode=forcefallback -Db_sanitize=thread ../..
             ninja src/btl/asyncstresstest
 
-        - name: Tests (TSan)
+        - name: Tests (async stress)
           run: |
-            # exitcode=66 makes a TSan-reported data race fail the job; do not
-            # halt on the first report so every racing site is shown.
-            TSAN_OPTIONS="halt_on_error=0 exitcode=66" ./build/tsan/src/btl/asyncstresstest
+            ./build/debug/src/btl/asyncstresstest
 
     build-win:
         name: "build-win (${{ matrix.compiler.name }})"
--- a/.github/workflows/build-workflow.yaml
+++ b/.github/workflows/build-workflow.yaml
@@ -85,7 +85,9 @@ jobs:
 
         - name: Tests (TSan)
           run: |
-            TSAN_OPTIONS="halt_on_error=1 exitcode=66" ./build/tsan/src/btl/asyncstresstest
+            # exitcode=66 makes a TSan-reported data race fail the job; do not
+            # halt on the first report so every racing site is shown.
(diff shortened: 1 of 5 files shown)


samienne:
Closing this as an unmerged investigation rather than merging it.

**Outcome captured:** the macOS flakiness that motivated this was primarily a *timing* artifact (cancellation racing a wall-clock sleep under CI load), not the diagnosed memory-ordering bug — and that timing flakiness is now fixed deterministically in #95 (manual-promise tests, no sleeps). The store-buffering `whenAll` race hypothesized here was never confirmed to actually fire on the Apple-Silicon runner (racy code showed 0 leaks at wide margins), and the spinlock fix on this branch was therefore never validated red→green.

**If the whenAll race is picked up later, do it as a fresh, clean PR:**
1. Add a *double-miss instrumentation counter* to confirm the bug is real on ARM64 (timing-independent), giving a true 0/reject signal.
2. Fix with an **atomic rendezvous** (single `fetch_add` to 2; the second arrival resets `values_`) rather than the spinlock — no double-miss, no double-reset, no lock. `merge.h` has the identical handshake.
3. Salvage the independently-correct `futurecontrol.h` TSan-annotation fix from this branch.

Branch is left intact for reference; nothing here should merge (it carries TEMP CI changes).

PR #95, opened 2026-07-17
btl: make async ordering tests deterministic without sleeps

samienne:
## What & why

Several tests in `src/btl/test/asynctest.cpp` used `std::this_thread::sleep_for`
as a stand-in for *"this task hasn't finished yet"* and then asserted on the
race outcome. Under CI load the sleep loses the race — `whenAllCancelOnFail` is
the source of the intermittent macOS flake — and the long sleeps everywhere made
the suite slow.

This is a **test-design** issue, not a library bug: the sleeps were an ordering
*proxy*. This PR makes the ordering **explicit** instead of timed, so the tests
are deterministic and fast. It is **test-code only** — no production changes.

## The helper

New test-only header `src/btl/test/manualfuture.h`:

```cpp
template <typename... Ts>
std::pair<future::Promise<Ts...>, future::Future<Ts...>> makeManualFuture();
```

It builds a `FutureControl<Ts...>` + weak `Promise` + owning `Future`, mirroring
what `async.h` does internally but **without the threadpool**. Keep the `Future`
alive to use the `Promise`. Being able to call `p.set(...)` / `p.setFailure(...)`
by hand is what makes the ordering explicit.

## Converted

- **`whenAllCancelOnFail`** — the canonical rewrite: fail input 1 via
  `p1.setFailure(...)`, assert the whenAll throws and the sibling continuation
  did not run, then complete input 2's upstream via `p2.set()` and re-assert.
  Cancellation is proven deterministically (the dropped downstream control makes
  the continuation's `weak_ptr` fail to lock) — no sleep, no 100-iteration loop.
- **`autoCancel`** — drop the downstream future before completing the upstream
  manually, then assert the continuation didn't run. Same weak-ptr mechanism,
  no 200 ms sleep.
- **`makeFuture` helper** and the **`then`-chain** test — removed sleeps that
  were pure slowdown; those tests are already deterministic via blocking `.get()`.

## Left alone (deliberately)

- **`async.perf`** — throughput benchmark, not a correctness race. Untouched
  (including its warmup sleep).
- **`async.delayed`** — genuinely tests the `TimedQueue`/`btl::delayed` timer and
  measures elapsed time; it needs a controllable clock (see follow-up). Untouched.
- **`whenAllFail`, `mergeFail`** — large-loop concurrency/failure stress tests
  that are *about* running many real async jobs. Left as real threaded stress
  (a parallel effort covers the actual whenAll race); not neutered into
  single-threaded.

## Verification (clang-cl 18.1.7, the CI compiler)

- Full `btl` suite: **66 tests pass**.
- Flake loop: `whenAllCancelOnFail` + `autoCancel` at `--gtest_repeat=3000` →
  **0 failures**, every repeat green.
- Speedup (btl suite, clang-cl debug):
  - `whenAllCancelOnFail`: **10888 ms → 0 ms**
  - `async` suite: **16934 ms → 2978 ms**
  - full btl suite: **16995 ms → 2983 ms** (~5.7× faster)

The residual ~3 s is dominated by the intentionally-untouched `mergeFail`
(~1.9 s) and `whenAllFail` (~0.7 s) stress tests.

## Follow-up: injectable executor / virtual clock

Making `async.delayed` and the throughput/concurrency stress tests both fast
*and* deterministic would require a **production** change: let `async()`/`then()`
accept an executor carrying a controllable virtual clock plus a pump
(`handleTimers`/`drain`) that tests advance manually. That is deferred by design
— this PR deliberately stays test-only via the manual-promise approach. Recorded
in `docs/decisions.md` and `src/btl/AGENTS.md`.

### DIFF
--- a/src/btl/test/asynctest.cpp
+++ b/src/btl/test/asynctest.cpp
@@ -1,22 +1,21 @@
 #include "manualfuture.h"
 
+#include <btl/always.h>
+#include <btl/async.h>
+#include <btl/delayed.h>
+#include <btl/future/future.h>
 #include <btl/future/join.h>
 #include <btl/future/merge.h>
-#include <btl/future/whenall.h>
 #include <btl/future/sharedfuture.h>
-#include <btl/future/future.h>
-
-#include <btl/delayed.h>
-#include <btl/async.h>
-#include <btl/reduce.h>
+#include <btl/future/whenall.h>
 #include <btl/plus.h>
-#include <btl/always.h>
+#include <btl/reduce.h>
 
 #include <gtest/gtest.h>
 
 #include <atomic>
-#include <random>
 #include <chrono>
+#include <random>
 #include <thread>
 #include <type_traits>
 
--- a/src/btl/test/asynctest.cpp
+++ b/src/btl/test/asynctest.cpp
@@ -1,4 +1,5 @@
-#include <atomic>
+#include "manualfuture.h"
+
 #include <btl/future/join.h>
 #include <btl/future/merge.h>
 #include <btl/future/whenall.h>
@@ -11,10 +12,9 @@
 #include <btl/plus.h>
 #include <btl/always.h>
 
-#include "manualfuture.h"
-
 #include <gtest/gtest.h>
 
+#include <atomic>
 #include <random>
 #include <chrono>
 #include <thread>
--- a/src/btl/test/asynctest.cpp
+++ b/src/btl/test/asynctest.cpp
@@ -11,6 +11,8 @@
 #include <btl/plus.h>
 #include <btl/always.h>
 
+#include "manualfuture.h"
+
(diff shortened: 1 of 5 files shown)


src/btl/test/asynctest.cpp
```diff
@@ -11,6 +11,8 @@
 #include <btl/plus.h>
 #include <btl/always.h>
 
+#include "manualfuture.h"
```
samienne: Put this in the top block. The order of the includes must be 1. the header that matches the .cpp if applicable, 2. the "local" headers that are part of the same lib/module. 3. Blocks are should be in dependency order. Everything with the least dependencies should come first. If lib a depends on lib b then a comes first. This only applies on per lib/block so not within the library it self. Last c++ and c headers, c++ first, c last.

PR #96, opened 2026-07-18
Add a JSON-RPC inspector protocol and headless platform

samienne:
A running app exposes an inspector protocol: a JSON-RPC 2.0 control channel an out-of-process client (an agent or an editor) launches, drives, and queries. The app is client-agnostic. See docs/design/inspector-protocol.md.

## Protocol
- JSON-RPC 2.0: id/method/params, result/error, id-less notifications both ways. nlohmann/json, confined to the agent-layer .cpp files.
- Launches PAUSED. Flow control: `app.run`, `app.pause`, `app.step`, `app.setFrameNotifications`, `app.shutdown`.
- Queries: `system.describe` (self-describing method registry), `window.list`, `window.introspect`, `window.renderTree`, `window.inject`.
- Windows addressed by `btl::UniqueId`.

## Execution
- A transport reader thread feeds a thread-safe command queue; the app thread stays single-threaded and drains at a frame boundary. Queries are coherent whether running or paused.
- Headless, frame-steppable platform (dummy backend), selected by env (`REACTIVE_HEADLESS` / `REACTIVE_PLATFORM=dummy`) or programmatically. Agentic mode is orthogonal.

## Transports
- Framed stream, 4-byte big-endian length prefix. Named pipe (Windows), unix socket (POSIX), TCP (`tcp://host:port`). App connects; client listens.

## Client bridge
- An LLM reaches the protocol through a separate MCP bridge (`tools/mcp/bqui_mcp.py`), never raw. One MCP tool per method, generated from `system.describe`.

bq untouched.


src/bqui/src/agent/transport.cpp
```diff
@@ -0,0 +1,373 @@
...
+        }
+
+        return true;
+    }
+
+private:
+    int fd_;
+};
+
+class SocketListener : public TransportListener
```
samienne: Let's split these so that there's only one class per cpp file (some small helpers might be ok to have in the same .cpp)

src/bqui/src/app.cpp
```diff
@@ -37,18 +39,61 @@
...
 namespace bqui
 {
 
+class AppDeferred;
+
 namespace
 {
+    // The app currently inside App::run, so the free withAnimation() can reach
+    // its live windows. Set for the duration of a run and cleared after.
+    AppDeferred* s_currentApp = nullptr;
```
samienne: Any globally accessible objects must be protected for threaded access.

src/bqui/include/bqui/app.h
```diff
@@ -22,6 +26,29 @@ namespace bqui
 
         App windows(std::initializer_list<Window> windows) &&;
 
+        /// Force a specific ase platform (e.g. a headless one), bypassing env
+        /// selection. Lets a test run headless without a process-global env var.
+        App platform(ase::Platform platform) &&;
+
+        /// Force headless (dummy) or headful, overriding the REACTIVE_HEADLESS /
+        /// REACTIVE_PLATFORM env vars.
```
samienne: wrong comment type

src/bqui/include/bqui/app.h
```diff
@@ -22,6 +26,29 @@ namespace bqui
 
         App windows(std::initializer_list<Window> windows) &&;
 
+        /// Force a specific ase platform (e.g. a headless one), bypassing env
+        /// selection. Lets a test run headless without a process-global env var.
```
samienne: wrong comment type

src/bqui/include/bqui/agent/session.h
```diff
@@ -0,0 +1,69 @@
...
+     * @brief Serve the agent over `transport` until it quits or disconnects.
+     *
+     * Synchronous request/response over the connected transport: receive one
+     * command, apply it to the windows, send back a snapshot of all of them,
+     * repeat. Commands are `step` (route each inject to its target window by
+     * index, then advance every window one dt), `snapshot` (observe without
+     * advancing), and `quit`. A malformed command is answered with an error,
+     * not a crash; a clean channel close ends the loop.
+     */
+    BQUI_EXPORT void runSession(std::vector<AgentWindow*> const& windows,
```
samienne: raw pointers here sound a bit code smell here. Can you elaborate why we need them?

src/bqui/include/bqui/agent/session.h
```diff
@@ -0,0 +1,69 @@
...
+            Transport& transport);
+
+    /** @brief Connect to `endpoint` and run a session over all `windows`. */
+    BQUI_EXPORT void runSession(std::vector<AgentWindow*> const& windows,
```
samienne: same here.

src/bqui/include/bqui/app.h
```diff
@@ -22,6 +26,29 @@ namespace bqui
...
+         * Orthogonal to the platform choice. In agentic mode the app connects to
+         * an endpoint and is driven by an external agent instead of free-running.
+         */
+        App agentic(bool agentic) &&;
+
+        /**
+         * @brief Set the agent channel endpoint, overriding
+         * REACTIVE_AGENT_ENDPOINT, so a test need not set a process-global env.
+         */
+        App agentEndpoint(std::string endpoint) &&;
```
samienne: I also think that the endpoint as a string needs some explanation what it expects.

src/ase/meson.build
```diff
@@ -29,6 +29,24 @@ glsrcs = [
 # a display and a GPU context that CI does not have.
 ase_is_headless = false
```
samienne: Is this needed anymore?

src/bqui/include/bqui/app.h
```diff
@@ -39,6 +42,35 @@ namespace bqui
...
+        App& headless(bool headless);
+
+        /** @brief Force agentic mode on/off, overriding the REACTIVE_AGENT /
+         * REACTIVE_MODE env vars.
+         *
+         * Orthogonal to the platform choice. In agentic mode the app connects
+         * to an endpoint and is driven by an external agent instead of
+         * free-running.
+         */
+        App& agentic(bool agentic);
```
samienne: Let's come up with some other name for this.

src/bqui/include/bqui/app.h
```diff
@@ -39,6 +42,35 @@ namespace bqui
...
+
+        /** @brief Set the agent channel endpoint, overriding
+         * REACTIVE_AGENT_ENDPOINT, so a test need not set a process-global env.
+         *
+         * A fully-local IPC endpoint the external agent listens on and the app
+         * connects to: a named-pipe name (`\\.\pipe\<name>`) on Windows, or a
+         * Unix-domain socket path on POSIX.
+         */
+        App& agentEndpoint(std::string endpoint);
```
samienne: Maybe this could be combined with the .agentic. Something like setRemoteEndpoint, or even addRemoteAddpoint if we are able to support multiple endpoints. This could be useful e.g. if we allow an agent to control the app and then I have a inspection ui connected at the same time. Adding an endpoint will put the app in "remote" mode automatically. Or maybe there is a single endpoint with multiple connections.

src/bqui/src/agent/session.cpp
```diff
@@ -0,0 +1,756 @@
...
+
+/**
+ * @brief The inspector-protocol server: a method registry and its dispatch loop.
+ *
+ * Holds the app seam, the monotonic frame counter, and the run state. Launches
+ * paused: run() starts a reader thread feeding a command queue, then runs a
+ * single-threaded state machine over {PAUSED, RUNNING} — the app thread is the
+ * only one that touches windows, sends responses, and emits notifications.
+ */
+class Session
```
samienne: Let's split into its own .cpp/.h

src/ase/meson.build
```diff
@@ -24,10 +24,23 @@ glsrcs = [
...
   ]
 
-# Whether the backend built here is the headless "dummy" one. A test that
-# drives a real window can only run where it is: the GLX and WGL backends need
-# a display and a GPU context that CI does not have.
-ase_is_headless = false
+# The headless (dummy) backend is compiled on every platform so any app can run
+# without an OS window; it has no GL/OS dependencies. makeDefaultPlatform() is
+# defined by the OS backend (wgl/glx), or by dummydefaultplatform.cpp when there
+# is none.
```
samienne: Remove the comment

src/bqui/include/bqui/app.h
samienne: Already addressed on the current branch: the triple-slash doxygen on platform() (and the rest of this block) are now /** */ block comments, so no wrong comment type remains here.

src/bqui/include/bqui/agent/session.h
samienne: Same as the app.h thread -- session.h now uses /** */ block doxygen throughout and no triple-slash remains. Already addressed.

src/bqui/include/bqui/app.h
samienne: Renamed. agentic(bool) is gone: it and agentEndpoint() are now the single App::setRemoteEndpoint(std::string) (ee183c3), and the whole layer is renamed bqui::agent -> bqui::remote.

src/bqui/include/bqui/app.h
samienne: Done: combined into App& setRemoteEndpoint(std::string endpoint). A non-empty endpoint puts the app in remote mode automatically; an empty override forces it off. I chose set rather than add because the session owns the app clock and blocks until shutdown, so a single endpoint/connection is what the model supports today -- multiple concurrent drivers would need a real multi-session reactor (the runloop rework), which is not here yet. The env var is now REACTIVE_REMOTE_ENDPOINT.

src/bqui/src/agent/session.cpp
samienne: Already addressed: the JSON adapter is its own introspectionjson.{h,cpp} translation unit, each transport is its own .cpp/.h, and the former reader-thread CommandQueue is gone. session.cpp now holds one primary class (Session) plus small local helpers, per the one-class-per-cpp guideline.

src/ase/meson.build
samienne: Removed (1c68f72).

src/bqui/src/windowimpl.h
```diff
@@ -444,6 +445,32 @@ class WindowImpl
...
+    ase::Window& getWindow()
+    {
+        return aseWindow;
+    }
+
+    /** @brief Advance this window by one deterministic frame on demand.
+     *
+     * Used by the agent to step frames synchronously. Forces the update
+     * regardless of the on-demand observe gate, so a stepped frame always
+     * consumes injected events and re-evaluates the signal graph. */
```
samienne: agent->remote. Also I'm wondering why we have to force the update. I was thinking that this should work just the same way as the normal loop but just controlled from the remote connection.

src/bqui/meson.build
```diff
@@ -68,44 +78,61 @@ libbquisrc = [
 
 bquidep = [libbtldep, libbqdep, libasedep, libavgdep]
 
+# nlohmann/json owns the remote layer's JSON-RPC envelope and serialisation. It
+# is header-only and confined to the remote layer's own .cpp files (json.hpp is
+# never pulled into a public header), so it is a dependency of the bqui library
+# only — deliberately *not* re-exported through libbquidep, keeping its compile
+# cost from spreading to dependents.
```
samienne: Remove comment

src/bqui/src/windowimpl.h
samienne: Done, both parts.

Rename (agent -> remote, outside the agent/ dir): windowimpl.h `agentTime_` -> `remoteTime_` and its uses; the getWindow()/stepFrame()/getResolvedIntrospection() doc comments and the clock comment now say "a remote driver" instead of "the agent"; app.h setRemoteEndpoint() drops "(an agent, an inspection UI)" for "(an inspection UI, an automation driver)"; and the accepted-connection local in the two real-app session tests is renamed `agent` -> `client` (matching the fixture). Left as genuinely unrelated: the AGENTS.md / docs "agent" usage, and the test-local literals (the "Agent" window title and the agentcore/agentdynamic endpoint names).

Force: you were right, it was a bypass and it is gone.

How the normal loop makes a frame: a real input event runs an ase window callback (e.g. buttonCallback_), which emits into an input area; the handler sets a bq input, and InputHandle::set() synchronously fires the observe callback registered on widgetInstanceSignal_ (the observe control is propagated onto every input dependency). That callback is the `wake` lambda -> needsUpdate_ = true, plus requestFrame(). onFrame() then does needsUpdate_.exchange(false) and re-evaluates in makeTransaction(); animation is a separate branch keyed on animating_.

An injected event takes the exact same path: GenericWindow::injectPointerButtonEvent (and the key/text/hover variants) call those same callbacks. So window.inject already sets needsUpdate_ through observe, and it stays set (it is a member) until the next app.step. stepFrame therefore only needs to advance remoteTime_ and call onFrame -- the injected event woke it, and it re-evaluates on any real change, uniform with the platform loop. Removed the `needsUpdate_ = true;` from stepFrame.

Test impact: none needed. The two real-app end-to-end tests that actually depend on a step re-evaluating after an inject -- session.describeListIntrospectDriveARealApp (counter 0 -> 1) and session.dynamicWindowsOpenAndCloseById (open/close child) -- both still pass without the force. The Session-level step tests (stepCountAdvancesAndFrameIsMonotonic, stepFromPausedAdvancesExactlyAndStaysPaused, pausedFrameIsStableWithoutAStep) assert on FakeRemoteWindow::advances() (the stepFrame call count) and the frame counter, not on re-evaluation, so they are unaffected. Full session suite 21/21; bqui 123/123; btl/bq/avg all green (clang-cl).

src/bqui/meson.build
samienne: Removed.

samienne:
Superseded by #142, which reworks the JSON-RPC inspector + remote-driven platform onto the merged #141 architecture. The agentic-platform approach here conflicts with #141 and carries nothing unique. Closing; branch retained.

PR #99, opened 2026-07-19
Make App's window list reactive

samienne:
## What

Makes `App`'s window list **reactive** — a signal of `(stable-id, Window)` pairs — so windows can be added and removed at runtime, mirroring how the toolkit already turns a mutable collection into a dynamic, stable-id widget list (`Collection` + `dataBind`).

## Window API

- **`WindowList = AnySignal<std::vector<std::pair<size_t, Window>>>`** — the reactive window list. The `size_t` is a producer-assigned stable id; the run loop reconciles by id. Same shape `dataBind` produces for widgets.
- **`App::windows(WindowList)`** — new overload consuming the signal directly. Per the scope note, there is no windows-specific `Collection`/`DataSource` layer; a caller produces the signal however it likes.
- **`App::windows(std::initializer_list<Window>)`** — unchanged signature; now enumerates the windows into a *constant* `WindowList` (ids `0..n-1`) so the static path flows through the same by-id machinery and behaves identically to before.
- **`App::onWindowsReconciled(callback)`** — observation hook reporting the live window id set after each reconciliation (headless introspection / testing).

## Reconciliation & lifecycle

`App::run` samples the window list once per frame via a dedicated `SignalContext`, at the top of the loop **before** the running signal is evaluated — a single well-defined reconcile point. It maintains `WindowGlue`s keyed by stable id:

- **new id → create** its `WindowGlue` (which constructs the `ase::Window` on demand via the platform + shared render context),
- **vanished id → destroy** its glue (closing the window),
- **existing id → keep** across frames.

The glue vector order follows the list. On WGL a mid-run `ase::Window` auto-registers in the platform's window map and is serviced automatically; on destruction its weak ref expires and is pruned. The no-arg `run()` still stops the app when any window closes — the onClose wiring is now a `.map` over the list so it stays reactive and applies to windows added mid-run.

## Behavior preservation

The static `initializer_list` path produces identical behavior/output to before (single reconcile at startup, contiguous ids from zero). `testapp1` builds and runs unchanged.

## Tests

New `src/bqui/test/apptest.cpp` (headless dummy platform):
- `dynamicWindowsReconcileById` — start with one window, add a second, then remove the first, driving edits from the reconcile observer; asserts the observed id sets are `[0] → [0,1] → [1]`.
- `staticWindowsEnumerateFromZero` — the `initializer_list` path enumerates ids `[0,1,2]`.

## Verification

Configured and built with **clang-cl** (`CC=clang-cl CXX=clang-cl` from the MSVC env) and ran the full `meson test` suite — all four suites (btl/bq/avg/bqui) pass, including the two new App tests. Clean-context style and correctness reviews both came back clean.

## Docs

Updated `src/bqui/readme.md` (Application section), `src/bqui/AGENTS.md` (new "app loop" section + verified-against stamp), and header Doxygen.

### DIFF
--- a/src/bqui/AGENTS.md
+++ b/src/bqui/AGENTS.md
@@ -1,6 +1,6 @@
 # bqui — agent notes
 
-*Last verified against `d2e8954` (2026-07-13).*
+*Last verified against `221cd3f` (2026-07-19).*
 
 Internals, entry points, and traps for the UI toolkit. Concepts and usage are in
 `readme.md`; project-wide conventions are in the top-level `docs/`. This file is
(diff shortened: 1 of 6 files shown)


samienne:
Superseded by #118, which reworks this onto `ArraySignal`.

The rework settles the design question this PR was waiting on — whether any identity surfaces at the exit. It does not: `App` takes an `ArraySignal<Window>` and builds one `WindowGlue` per identity with `map()`, so `WindowList`, the producer-assigned `size_t` and the hand-written by-id reconcile loop are all deleted. The consumer's `map` *is* the reconcile.

For the record, the CI failures here were never staleness: `App::run` opens real windows and the runners have no display or GPU context (`Unable to open X display`, `Unable to get wglCreateContextAttribsARB`). #118 adds an `ase_is_headless` export so the app tests compile only where they can run.

PR #100, opened 2026-07-19
ArraySignal<T>, forEach, and the keyed layout operators

samienne:
Design record plus the `bq` implementation it specifies.

## What is here

**The design** — `docs/design/arraysignal.md`. `ArraySignal<T>` as a value in
`bq` describing a list whose contents and membership may both change; `forEach`
as the keyed operator that builds once per identity; `KeyedArraySignal<A>` with
`extract`/`scatter` as the fan-in/fan-out pair a container needs. The headline
finding is that `bqui`'s two layout engines (`layout()` and `dynamicBox`)
collapse into one — `layout()`'s own `ObbMap`/`SizeHintMap` aliases already *are*
this API.

**Steps 1 and 2 of it, built.** Two new public headers in `bq`:

- `bq/signal/arraysignal.h` — `ArraySignal<T>`, its constructors, `map`,
  `forEach`, `concat`.
- `bq/signal/arraysignallayout.h` — `KeyedArraySignal<A>` with `mapValues`,
  `mapKeyed` and `values`, plus the free functions `extract` and `scatter`, in
  `bq::signal::layout`.

22 tests in `src/bq/test/arraysignaltest.cpp`. **Nothing in `bqui` changes** —
`layout()` and `dynamicBox` are untouched; steps 3 and 4 are separate work.

## The part that matters

Once-per-identity initialisation is what stops a sibling insert from resetting
every surviving child. `Join::update` re-initialises its inner signal whenever
the outer changes, which discards everything held in the positional `DataType`
tree below it — `map` caches, `withPrevious`, `iterate` folds. The fan-in is
still `combine(...).join()`; what makes it safe is that every per-identity signal
is `.share()`d, because `SharedControl` keys its per-context state by a
`btl::UniqueId` in the `DataContext` and so re-finds it instead of building
fresh. On top of that, element equality compares identity only, so `tryCheck()`
on an element sequence means "membership unchanged" and a value change never
enters the re-initialisation path at all.

`forEach`, `map`, `extract` and `scatter` all run through **one**
apply-once-per-identity operation, so they cannot drift apart the way `layout()`
and `dynamicBox` did.

**Mutation-verified.** Dropping the single `.share()` in `extract` turns exactly
`siblingStateSurvivesInsertionAndRemoval` red and nothing else. Making the
once-per-identity cache rebuild on every emission turns five red, including both
`forEach` delegate-count tests.

## Two pre-existing `bq` bugs fixed on the way

`join()` did not work on type-erased signals, in two independent ways:

- `Signal::join()` instantiated `Join<TStorage>`, and `TStorage` is `void` for an
  `AnySignal` — a hard error.
- `detail::toSignal`'s forwarding overload beat its `Signal<T, Ts...>` overload
  for an `AnySignal` (exact match vs derived-to-base), so `join()` over a signal
  whose *value* was an `AnySignal` wrapped it in a `constant` and **silently did
  not join**.

## Where the design did not survive contact

Recorded in the doc under *What building it changed*, in full. The substantive
ones:

- **`extract`'s function takes `AnySignal<T>`, not `T`.** A `T`-taking `f`
  cannot be applied once per identity — reaching a `T` means evaluating the
  element's signal, which needs a `SignalContext` the operator does not have,
  and `value.map(f).join()` is precisely the `Join` hazard.
- **`mapValues` passes its extras first**, resolving a sentence that
  contradicted itself: the claim that it matches `ObbMap` "down to the argument
  order" is the load-bearing half, since it is what lets an existing `ObbMap` be
  passed straight in.
- **Ids are namespaced by allocator identity**, so `concat` rewrites nothing —
  but ids embed a pointer and are therefore not reproducible between runs. That
  retires the doc's argument that a test can assert on the id set. No loss: ids
  never surface.
- **The duplicate-key release rule is settled** by occurrence index. Every
  occurrence keeps a stable identity; nothing is dropped and nothing churns.
  Strictly better than the candidate rule it supersedes.

Also corrected, from PR #103's empirical result: the claim that `getSizes`'
divide-by-zero is masked only by `std::min`'s argument order is **wrong** and is
dropped. The reversed-operand mutation leaves all 24 layout tests green, because
a zero combined interval means every child's interval at that index is zero too.
The non-monotonic overflow hazard stands and is now pinned by a test.

## Deliberately not done

`join`/`flatMap` and `filter` (nothing needs them; `filter`'s eviction semantics
are an open question), the `AnySignal<ArraySignal<T>>` reactive-subtree
constructor (the doc does not yet agree with itself about whether it mints
identity), `map` with extra signals (waits on `Signal::map` gaining that form
first, as the design requires), and the defaulted `forEach` delegate.

The open question about whether identity surfaces at the exit is untouched —
that is PR #99's call and it does not block the core.

## Verification

Full tree configured and built with `clang-cl`, debug. All four suites pass
(`bq`, `btl`, `avg`, `bqui`).

### DIFF
--- a/src/bq/include/bq/signal/arraysignal.h
+++ b/src/bq/include/bq/signal/arraysignal.h
@@ -1,0 +1,634 @@
+#pragma once
+
+#include "combine.h"
+#include "constant.h"
+#include "join.h"
+#include "merge.h"
+#include "signal.h"
+
+#include <atomic>
+#include <cassert>
+#include <cstddef>
+#include <cstdint>
+#include <functional>
+#include <initializer_list>
+#include <iterator>
+#include <map>
+#include <memory>
+#include <mutex>
+#include <set>
+#include <type_traits>
+#include <utility>
+#include <vector>
+
+namespace bq::signal
+{
+    template <typename T>
+    class ArraySignal;
+
+    namespace detail
+    {
+        /** @brief The identity space of one ArraySignal.
+         *
+         * Constructed together with the ArraySignal that owns it. Ids are
+         * unique within one space and mean nothing outside it, and composing
+         * two arrays keeps their spaces distinct — which is what makes their
+         * ids distinct without any shared counter.
+         */
+        class ArrayIdSpace
+        {
+        public:
+            uint64_t allocate()
+            {
+                return next_.fetch_add(1, std::memory_order_relaxed);
+            }
+
+        private:
+            std::atomic<uint64_t> next_{ 1 };
+        };
+    } // namespace detail
+
+    /** @brief The identity of one element of one ArraySignal.
+     *
+     * Ids never reach user code: they are minted by an ArraySignal and matched
+     * within it. An id is equality-comparable and orderable so that it can be
+     * used as a map key; nothing else about it is meaningful, and in particular
+     * its ordering says nothing about element order.
+     */
+    class ArrayId
+    {
+    public:
+        ArrayId() = default;
+
+        ArrayId(std::shared_ptr<detail::ArrayIdSpace> space, uint64_t index) :
+            space_(std::move(space)),
+            index_(index)
+        {
+        }
+
+        bool operator==(ArrayId const& rhs) const
+        {
+            return index_ == rhs.index_ && space_ == rhs.space_;
+        }
+
+        bool operator!=(ArrayId const& rhs) const
+        {
+            return !(*this == rhs);
+        }
+
+        bool operator<(ArrayId const& rhs) const
+        {
+            if (space_ != rhs.space_)
+                return std::less<void const*>()(space_.get(), rhs.space_.get());
+
+            return index_ < rhs.index_;
+        }
+
+    private:
+        std::shared_ptr<detail::ArrayIdSpace> space_;
+        uint64_t index_ = 0;
+    };
+
+    namespace detail
+    {
+        /** @brief One realised element: its identity and its value over time.
+         *
+         * An element that survives a membership change keeps both — in
+         * particular the very same value signal object, so anything an operator
+         * built on that signal keeps its state.
+         *
+         * Two elements are equal when they have the same identity; the value is
+         * not part of the comparison. That makes an element sequence
+         * equality-comparable, so `tryCheck()` on it suppresses a change
+         * notification exactly when membership is unchanged.
+         */
+        template <typename T>
+        struct ArrayElement
+        {
+            ArrayId id;
+            AnySignal<T> value;
+
+            bool operator==(ArrayElement const& rhs) const
+            {
+                return id == rhs.id;
+            }
+
+            bool operator!=(ArrayElement const& rhs) const
+            {
+                return id != rhs.id;
+            }
+        };
+
+        template <typename T>
+        using ArrayElements = std::vector<ArrayElement<T>>;
+
+        /** @brief Applies a function once per identity and caches the result.
+         *
+         * This is the shared primitive under `map`, `forEach`, `extract` and
+         * `scatter`. The cache is keyed by identity rather than by position, so
+         * a sibling insertion or a reorder does not disturb it. An entry is
+         * evicted when its id leaves the array, and a re-appearing id is a new
+         * item that is built again.
+         *
+         * The cache lives with the operator rather than in a SignalContext,
+         * because everything it holds is an immutable description — a signal,
+         * or a value the delegate built. It is mutex-guarded because a signal
+         * may be evaluated from any thread.
+         */
+        template <typename T, typename U, typename TFunc>
+        class OncePerId
+        {
+        public:
+            OncePerId(TFunc func) :
+                func_(std::move(func))
+            {
+            }
+
+            ArrayElements<U> apply(ArrayElements<T> const& elements)
+            {
+                std::lock_guard<std::mutex> lock(mutex_);
+
+                ArrayElements<U> result;
+                result.reserve(elements.size());
+
+                std::set<ArrayId> live;
+                for (auto const& element : elements)
+                {
+                    auto i = cache_.find(element.id);
+                    if (i == cache_.end())
+                    {
+                        i = cache_.emplace(element.id,
+                                func_(element.id, element.value)).first;
+                    }
+
+                    live.insert(element.id);
+                    result.push_back({ element.id, i->second });
+                }
+
+                for (auto i = cache_.begin(); i != cache_.end();)
+                    i = live.count(i->first) ? std::next(i) : cache_.erase(i);
+
+                return result;
+            }
+
+        private:
+            std::mutex mutex_;
+            std::map<ArrayId, AnySignal<U>> cache_;
+            TFunc func_;
+        };
+
+        /** @brief Runs `func` once per identity over an element signal. */
+        template <typename U, typename T, typename TFunc>
+        AnySignal<ArrayElements<U>> mapElementsOnce(
+                AnySignal<ArrayElements<T>> elements, TFunc func)
+        {
+            auto once = std::make_shared<OncePerId<T, U, TFunc>>(
+                    std::move(func));
+
+            return elements.map([once](ArrayElements<T> const& es)
+                {
+                    return once->apply(es);
+                });
+        }
+
+        /** @brief Fans the element values in, discarding identity. */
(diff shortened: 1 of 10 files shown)


samienne:
## Update: newly settled design folded in

Second commit (`dead4d9`). Still doc-only.

**Resolved two open questions.** Ids are assigned **at the exit**, by an allocator that is a parameter of the operation constructing the exit signal — promoted from working assumption to decision. This removes a real tension: a consumer-owned allocator and a per-type constructor set cannot both be satisfied at construction time, since nested `DynArray`s are built before their parent and the parent before it reaches any consumer. `forEach`'s key→id map follows the allocator's lifetime, so two exits over one `DynArray` legitimately get independent id spaces.

**New: `DynArray` as a domain alongside `Signal`.** `pure`, concatenation as a monoid (its associativity law *is* the brace-nesting ergonomics, `{a, {b, c}} ≡ {a, b, c}`), `join` — already latent in the recursive tree — `map`, `filter`, `flatMap`. Includes the **identity-algebra table** (create / preserve / namespace / reveal) as the practical guarantee, and records the boundary asymmetry: entering requires a key because the `Signal` domain carries no identity; leaving is free because the ids already exist.

**New rationale: `dynamicBox` is membership-only.** It caches by id and consults the cache before looking at the widget (`dynamicbox.h:113-124`), so when an existing item's **widget changes** the new builder is discarded and the change is **silently ignored**. It cannot do better — the only alternative is a rebuild, which destroys state. `forEach`'s `(AnySignal<T>) -> U` closes the hole rather than choosing between the two failures. **This makes the work a fix, not a refactor.**

**Also added:** the proposal that `DynArray` own a generic *apply-once-per-id* operation (`dynamicBox`, `App`'s window reconcile, and `forEach` are three copies of it — layering holds, since "apply once per identity" is generic); the naming split `join`/`flatten` vs `toSignal(allocator)`; the `.map` caveat (value-level and fine, vs the structure-level build delegate that is banned — same signature, different role, indistinguishable to the type system); `map` with extra signals; and a C++20 `==`/`<=>` note that the C++17 design must not depend on.

### Verified: `Signal::map` / `combine` today

- **`Signal::map` takes only a function** — `map(TFunc&&)`, `const&` and `&&` overloads (`signal.h:225-237`). **No extra-signals form exists.**
- `combine` is homogeneous and vector-shaped: `combine(vector<AnySignal<T>>) → AnySignal<vector<T>>` (`combine.h:73`).
- The heterogeneous fuse is `merge`, variadic, free (`merge.h:104`) and member (`signal.h:307-310`). So `a.merge(b, c).map(f)` is the proposed sugar, spelled out.

Since `DynArray::map` must mirror `Signal::map` exactly, this argues for **adding the extra-signals form to `Signal::map` first** — otherwise the two domains diverge exactly where the framing promises they agree.

### Two new open questions the framing surfaced

1. **The reactive subtree looks like a third minting site.** The table says only `pure` and `forEach` mint identity, but `AnySignal<DynArray<T>>` hands out fresh ids on every swap. Either it needs its own row, or the swap should be modelled as `join` over a signal-of-`DynArray` with fresh ids coming from re-entering the new subtree's `pure`s. The second is tidier and probably intended, but it is not what the wording currently says.
2. **`filter` and the apply-once-per-id cache disagree about eviction.** Filter an item out, then back in: upstream `forEach` still holds its id (it never left the source array), but the cache evicted the built value and "a re-appearing key rebuilds." So the delegate re-runs for an id that was never retired — contradicting "invoked only when a new id is introduced," and silently losing a filtered-out widget's state.

Also flagged inline: "construction assigns no ids" and "`map` preserves ids" are only consistent if identity inside the domain is **symbolic** (a tree position or key) and `toSignal` resolves it to a `size_t`. That is clearly the intended reading, so I stated it rather than leaving it as an apparent contradiction.

Other open questions left open as instructed: defaulted-delegate return type, duplicate-key release rule (kept "first occurrence keeps the id" as an explicit *candidate*, not adopted), `{x}`, and `Collection`/`DataSource` retirement.

src/bq/include/bq/signal/arraysignal.h
```diff
@@ -0,0 +1,634 @@
...
+        ArraySignal(AnySignal<std::vector<T>> values) :
+            ArraySignal(positionalElements(std::move(values)))
+        {
+        }
+
+        /** @brief Constructs a single constant item. */
+        template <typename U, typename = std::enable_if_t<
+            std::is_convertible_v<U, T>
+            && !detail::IsArraySignal<std::decay_t<U>>::value
+            && !detail::IsAnySignal<std::decay_t<U>>::value
```
samienne, line 322: I think std::is_convertible_v already handles both IsArraySignal and IsAnySignal because those cannot be converted to T anyways. Also I think it should be possible to have ArraySignal<AnySignal<T>> as ArraySignal should be able to handle pretty much any value type.

src/bq/include/bq/signal/arraysignal.h
```diff
@@ -0,0 +1,634 @@
...
+         *
+         * `keyFn` runs for every item on every change. A changed key is a new
+         * item: the old key's result is destroyed and the new key's is built.
+         *
+         * Duplicate keys assert in debug. In release each occurrence of a
+         * repeated key gets its own identity, in order of appearance, so
+         * nothing is dropped and nothing churns.
+         */
+        template <typename TKeyFunc, typename TDelegate>
+        auto forEach(TKeyFunc keyFn, TDelegate delegate) const;
```
samienne, line 367: TKeyFunc and TDelegate should be restricted with std::is_invocable_r or std::function.

src/bq/include/bq/signal/arraysignal.h
```diff
@@ -0,0 +1,634 @@
...
+        ElementsSignal elements_;
+    };
+
+    /** @brief Keys a changing list and builds one result per key.
+     *
+     * The entry into the ArraySignal domain, and the only operation other than
+     * constructing a single item that mints identity. See
+     * `ArraySignal::forEach` for the semantics of `keyFn` and `delegate`.
+     */
+    template <typename T, typename TKeyFunc, typename TDelegate>
```
samienne, line 503: TKeyFunc and TDelegate should be restricted with std::is_invocable_r

src/bq/include/bq/signal/arraysignal.h
```diff
@@ -0,0 +1,634 @@
...
+    auto forEach(AnySignal<std::vector<T>> source, TKeyFunc keyFn,
+            TDelegate delegate)
```
samienne, line 505: delegate should default to identity function.

src/bq/include/bq/signal/arraysignal.h
```diff
@@ -0,0 +1,634 @@
...
+
+                        for (auto const& entry : current.order)
+                        {
+                            auto i = state->items.find(entry);
+                            if (i == state->items.end())
+                            {
+                                T initial = current.byEntry.at(entry);
+
+                                auto value = AnySignal<T>(keyed
+                                        .map([entry, initial](
```
samienne, line 560: This will store the initial value indefinitely inside the signal.

src/bq/include/bq/signal/arraysignal.h
```diff
@@ -0,0 +1,634 @@
...
+            std::mutex mutex;
+            std::shared_ptr<detail::ArrayIdSpace> space =
+                std::make_shared<detail::ArrayIdSpace>();
+            std::map<Entry, detail::ArrayElement<T>> items;
+        };
+
+        auto state = std::make_shared<State>();
+
+        auto elements = AnySignal<detail::ArrayElements<T>>(
+                keyed.map([state, keyed](Keyed const& current)
```
samienne, line 545: This is quite dangerous to store the state outside of the signal context. This could be a perfect use case of signal::withPrevious.

samienne:
Superseded by #110, which carries the revised design.

The implementation here is a **spike** and is deliberately not being extended: it kept the per-identity table in a `shared_ptr` captured by a `map` lambda, so the state lived in the signal *description* rather than the `DataContext`. Two `SignalContext`s therefore shared one table and evicted each other's entries — and those entries are widget state, so it is a correctness defect rather than a limitation.

What was worth keeping has been carried forward: the two `bq` `join` bugs it found (both landed in #112 with their own tests) and its test cases. The design document itself is revised on #110.

PR #101, opened 2026-07-19
Fix use-after-move in uniformGrid and remove dead layout code

samienne:
Three independent fixes in the `bqui` layout code, one per commit so any of
them can be dropped on its own.

## 1. `uniformGrid` no longer moves out of its children

`UniformGrid::operator AnyWidget()` built a local `builders` vector by
invoking every element of `widgets` as an rvalue (`std::move(widget)(params)`),
then ran a loop that computed a per-cell scaled size hint into `hi` and threw
it away. Neither `builders` nor `hi` was read afterwards. The same `widgets`
vector was then forwarded to `layout()`, i.e. after every element had been used
as a move source.

Both the vector and the loop are dead, so they are simply deleted; `widgets`
now reaches `layout()` untouched. (`params` becomes unused in the build lambda,
so its name is dropped to avoid a new unreferenced-parameter warning.)

**On the observable symptom** — I traced what a moved-from `AnyWidget`
actually is before assuming a crash, and the honest answer is *not* a crash:

- `AnyWidget` is `Widget<std::function<AnyBuilder(BuildParams)>>` holding a
  `btl::CloneOnCopy<std::function<...>>`.
- `Widget::operator()(BuildParams) &&` calls
  `std::invoke(std::move(*func_), std::move(params))`. `std::function`'s call
  operator is `const`, so invoking it through an rvalue does *not* consume it;
  the target survives.
- The target is a `btl::ArgumentBinder`, whose `operator()` is also `const` and
  passes its bound state on via `btl::clone`, so no bound argument is moved
  out either.

So in the current code the "moved-from" state is a formality and the children
are still valid when `layout()` receives them. The concrete cost is that every
child widget is built twice — once into the discarded `builders`, once inside
`layout()` — with the first build's signal graph created and dropped. The
pattern is nonetheless a genuine use-after-move by contract: it only stays
benign because `AnyWidget`'s type erasure happens to be `const`-callable, and
any change to `Widget::operator() &&` that actually consumed `func_` (as
`operator AnyWidget() &&` already does) would turn this into a
`std::bad_function_call` on every child.

**Known limitation, deliberately left unfixed:** the deleted loop looks like a
vestige of an intended feature — scaling each child's size hint down by its own
`cell.w`/`cell.h` before aggregation. Today `mapHints` only scales the
aggregate by the grid dimensions (`multiplySizeHint(stackSizeHints(hints), w,
h)`), so a child spanning multiple cells does not influence the container's
size hint in proportion to its span. Implementing that is a behaviour change
and out of scope here.

## 2. `dynamicBox`: unused hint signal, and a missing `break`

(a) The function built two size-hint signals from `builders`: `hintsWithoutId`
and an id-tagged `hints`. `hints` is never read — `resultHint`, the
`makeWidgetWithSize` body and its argument list all use `hintsWithoutId`. The
binding (and the `combine(...).join().share()` graph it creates) is deleted.

(b) The `withPrevious` element cache scanned `previous` for a matching id but
did not `break` after moving the match into `result`. A repeated id would push
the same cached element twice, and the scan continued over a container from
which an element had already been moved out. Added the `break`.

## 3. `box.h`: unused width hints in `combineSizes`

The `Axis::y` branch of `combineSizes` computed `xHintResults` (every child's
`getWidth()`) and never used it; only `yHintResults` feeds `getSizes`. Deleted.

## Noticed but not fixed

- `uniformgrid.cpp` still includes `<bq/signal/combine.h>`, which was needed by
  the deleted `merge(...)` call. Left in place to keep the diff to the fixes.
- Neither `uniformGrid` nor `dynamicBox` has any test coverage; the duplicate-id
  path in (2b) in particular is unexercised.

## Verification

Configured and built with clang-cl (`CC=clang-cl CXX=clang-cl`, debug):

```
[241/241] Linking target src/bqui/bquitest.exe
1/4 reactive:bq   OK                0.16s
2/4 reactive:avg  OK                0.24s
3/4 reactive:bqui OK                0.20s
4/4 reactive:btl  OK                1.67s

Ok:                4
Fail:              0
```

No new warnings from the changed files (the remaining warnings are the
pre-existing Eigen `std::result_of` deprecations).

### DIFF
--- a/src/bqui/src/widget/uniformgrid.cpp
+++ b/src/bqui/src/widget/uniformgrid.cpp
@@ -1,14 +1,9 @@
 #include "bqui/widget/uniformgrid.h"
 
-#include "bqui/provider/providebuildparams.h"
-
 #include "bqui/widget/layout.h"
 
 #include "bqui/mapsizehint.h"
 #include "bqui/stacksizehint.h"
-
-#include <bq/signal/combine.h>
-#include <bq/signal/signal.h>
 
 namespace bqui::widget
 {
@@ -48,26 +43,9 @@ auto multiplySizeHint(SizeHint const& sizeHint, float x, float y) -> SizeHint
 
 UniformGrid::operator AnyWidget() &&
 {
-    return makeWidget([](BuildParams const& params, auto widgets, auto cells,
+    return makeWidget([](auto widgets, auto cells,
                 unsigned int w, unsigned int h)
         {
-            std::vector<AnyBuilder> builders;
-
-            for (auto&& widget : widgets)
-                builders.push_back(std::move(widget)(params));
-
-            size_t i = 0;
-            for (auto const& cell : cells)
-            {
-                auto hi = merge(
-                        builders[i].getSizeHint(),
-                        bq::signal::constant(1.0f / (float)cell.w),
-                        bq::signal::constant(1.0f / (float)cell.h)
-                        ).map(multiplySizeHint);
-
-                ++i;
-            }
-
             auto mapHints = [w, h](std::vector<SizeHint> const& hints)
                 -> SizeHint
             {
@@ -108,7 +86,6 @@ UniformGrid::operator AnyWidget() &&
                     );
 
         },
-        provider::provideBuildParams(),
         std::move(widgets_),
         std::move(cells_),
         w_,
--- a/src/bqui/include/bqui/dynamicbox.h
+++ b/src/bqui/include/bqui/dynamicbox.h
@@ -52,25 +52,6 @@ namespace bqui
                             return combine(std::move(hintSignals));
                     }).join().share();
 
-                auto hints = builders.map([]
-                        (std::vector<std::pair<size_t, widget::AnyBuilder>> const& builders)
-                    {
-                            std::vector<bq::signal::AnySignal<std::pair<size_t, SizeHint>>>
-                            hintSignals;
-                            for (auto const& builder : builders)
-                            {
-                                hintSignals.push_back(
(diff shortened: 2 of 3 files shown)


src/bqui/src/widget/uniformgrid.cpp
```diff
@@ -48,26 +48,9 @@ auto multiplySizeHint(SizeHint const& sizeHint, float x, float y) -> SizeHint
 
 UniformGrid::operator AnyWidget() &&
 {
-    return makeWidget([](BuildParams const& params, auto widgets, auto cells,
+    return makeWidget([](BuildParams const&, auto widgets, auto cells,
```
samienne: We can drop the provider for the BuildParams if they are not needed.

src/bqui/src/widget/uniformgrid.cpp
samienne: Right here.

samienne:
Addressed: dropped the `BuildParams` parameter and the `provider::provideBuildParams()` argument entirely, so the lambda is now `(widgets, cells, w, h)`. Confirmed `layout()` requests its own build params (`src/bqui/src/widget/layout.cpp:96`), so uniformGrid's outer widget never needed them.

That also made three includes dead, now removed: `bqui/provider/providebuildparams.h` (the provider), and `<bq/signal/combine.h>` / `<bq/signal/signal.h>` (the deleted `merge`/`bq::signal::constant` calls). `mapsizehint.h` and `stacksizehint.h` are still used.

Folded into the first commit and force-pushed. Rebuilt with clang-cl: 4/4 suites OK.

PR #102, opened 2026-07-19
bqui: test layout geometry through realised widgets

samienne:
Layout geometry currently has no test coverage. With a rewrite of the layouts
ahead, this establishes a harness that pins down *observable* geometry —
"given these children with these hints, in a box of this size, each child ends
up here with this size" — so it survives a reimplementation rather than
encoding today's internals.

## Approach: probe widgets read back through input areas

A probe is an empty widget with a fixed size hint that tags its realised
`Instance` with an `InputArea` carrying a unique id. Input areas are the one
geometry channel that travels intact from a leaf to the root `Instance`:
`Instance::transform` folds every enclosing transform into the area's own
transform while leaving its obb in leaf-local coordinates, and `addWidgets`
concatenates children's areas into the parent. Realising the layout once and
looking the areas up by id therefore recovers **both** the size a child was
allocated (the obb) and where it was placed (the transform), in the layout's
coordinates.

This matters because the geometry a child actually receives is decided
*outside* the child: `layout()` wraps each widget in `handleGravity()` and then
attaches `transformBuilder()` to the resulting builder. A probe that only
inspected itself would see its size but never its position. The input-area
channel sidesteps that, and has the bonus that it is the same data hit-testing
uses — asserting on it asserts something users depend on.

The tests name only `hbox`, `vbox`, `mapObbs`, `AnyWidget`, `SizeHint` and
`Instance`. Nothing about how the layouts are assembled internally is baked in.

### Alternatives considered

- **Walking the realised render tree.** More end-to-end, but obbs on
  `ContainerNode`/`ShapeNode` are local and would need flattening by hand, and
  the probes would have to draw something to appear at all. Strictly more
  machinery for the same numbers.
- **Testing `ObbMap`/`SizeHintMap` in isolation.** Cheap and precise, but it
  only covers the pure placement maths — not gravity, not the size-hint
  aggregation, not the transform plumbing that turns an obb into a child's
  actual position. Included here as a *complement* (the two `mapObbs` tests),
  not as the answer: the mutation run below shows exactly where they stop.
- **Introspection (PR #89).** Its window-space obb tree is close to
  purpose-built for this and would make the probe unnecessary — the assertions
  would read against a named tree instead of ids. It is unmerged, so this works
  on master without it; if #89 lands, `ProbeSet::realise` is the single place
  that would be re-pointed at the introspection tree, and the tests themselves
  would not change.
- **Headless `ase` "dummy" backend / real app loop.** Overkill. Layout is
  resolved entirely in the signal graph before anything is drawn; no window,
  GL context or frame loop is needed to observe it.

## SignalContext

The whole tree is realised and evaluated through **one** `SignalContext`:

```cpp
auto instanceSignal = std::move(widget)(BuildParams())(
        bq::signal::constant(size)).getInstance();

auto context = bq::signal::makeSignalContext(std::move(instanceSignal));
Instance const& instance = context.evaluate<0>().get<0>();
```

Contexts are parallel instantiations of the graph, so a signal peeked at in a
throwaway context is a *different* run of the same graph and diverges from the
live one (stream history is not replayed). Where a test needs two values, both
signals must go into the same `makeSignalContext(...)` call, addressed by
index — never a second context. Layout signals here are pure functions of
constant inputs, so a single `evaluate` with no `update()` is enough; anything
animated would need `update()` driven on that same context.

## Coverage

Covered: `hbox` filler distribution (partial and full), `vbox` reversed y axis,
`handleGravity` centering in an oversized slot, upward `SizeHint` aggregation
via `AccumulateSizeHint`, and `mapObbs` on both axes.

Not covered: `stack`, `uniformGrid`, `dynamicBox` (and its separate `MapObbs`
functor), nesting, zero/one-child and degenerate-hint edge cases, non-default
gravity, and animated/changing hints. All of those are additional `TEST`s
against the same `ProbeSet` — `stack` and `uniformGrid` need nothing new;
`dynamicBox` needs the probe list to be built from a signal; changing hints
need `context.update()` between evaluations, which is the one place the harness
would grow.

If breadth grows past this file, `ProbeSet` lifts to a `test/` header unchanged.

## Verification

Built and run with clang-cl (debug):

```
[==========] Running 7 tests from 1 test suite.
[ RUN      ] Layout.hboxDistributesFillerSpace
[       OK ] Layout.hboxDistributesFillerSpace (2 ms)
[ RUN      ] Layout.hboxGrantsFullFillerWhenSpaceAllows
[       OK ] Layout.hboxGrantsFullFillerWhenSpaceAllows (1 ms)
[ RUN      ] Layout.vboxStacksFromTopDown
[       OK ] Layout.vboxStacksFromTopDown (1 ms)
[ RUN      ] Layout.gravityCentersAChildInsideItsSlot
[       OK ] Layout.gravityCentersAChildInsideItsSlot (0 ms)
[ RUN      ] Layout.hboxAggregatesChildSizeHints
[       OK ] Layout.hboxAggregatesChildSizeHints (0 ms)
[ RUN      ] Layout.mapObbsPlacesChildrenLeftToRight
[       OK ] Layout.mapObbsPlacesChildrenLeftToRight (0 ms)
[ RUN      ] Layout.mapObbsPlacesChildrenTopToBottom
[       OK ] Layout.mapObbsPlacesChildrenTopToBottom (0 ms)
[  PASSED  ] 7 tests.
```

The full `bquitest` binary passes all 30 tests.

### The tests were checked against broken code

Two mutations were injected, built, run, and reverted.

**1. Off-by-one on the x accumulator (`acc += s[0] + 1.0f`) plus a reversed y
accumulator in `mapObbs`.** 5 of 7 red:

```
[  FAILED  ] Layout.hboxDistributesFillerSpace
[  FAILED  ] Layout.hboxGrantsFullFillerWhenSpaceAllows
[  FAILED  ] Layout.vboxStacksFromTopDown
[  FAILED  ] Layout.mapObbsPlacesChildrenLeftToRight
[  FAILED  ] Layout.mapObbsPlacesChildrenTopToBottom
```

with e.g. `Expected 40.0f ... Which is: 41` and, for the reversed axis,
`Expected 110 ... Which is: 150`. Importantly the *end-to-end probe* tests went
red alongside the direct `mapObbs` ones — the harness sees the perturbation
through the whole realisation chain, not just the pure function.

**2. Doubled gravity offset in `handleGravity`.** Exactly one test red, and only
the one that covers gravity:

```
[  FAILED  ] Layout.gravityCentersAChildInsideItsSlot
  Expected: y  Which is: 35
  Actual: geometry.position[1]  Which is: 70
```

Note this mutation left both `mapObbs` tests green — the isolated-maths tests
alone would have missed it, which is the concrete argument for the probe
harness over approach 3 on its own.

## Notes for review

- No bug was found in the layouts while writing these. Every expectation was
  computed by hand from `getSizes`/`mapObbs`/`handleGravity` before running, and
  all seven matched the implementation on the first run. Two fragilities are
  worth recording but are **not** touched here: `getSizes` divides by
  `combined[i] - combined[i-1]`, which is zero whenever consecutive hint
  components are equal (a very common case); the result is `inf` or `NaN`, and
  it clamps to a correct multiplier only because `std::min(1.0f, NaN)` returns
  `1.0f`. That is an accident of `std::min`'s argument order, not a designed
  invariant, and a rewrite should handle the degenerate interval explicitly.
  Also, `getSizes` does not clamp its output when `size` exceeds the total
  filler, so an oversized box overflows its children past its own bounds rather
  than leaving slack.
- `src/btl/test/asynctest.cpp` fails on this machine under debug clang-cl
  (`async.perf`, "can't increment invalidated vector iterator"). Pre-existing
  and unrelated — this PR adds no btl code.

### DIFF
--- a/src/bqui/meson.build
+++ b/src/bqui/meson.build
@@ -88,6 +88,7 @@ libbquidep = declare_dependency(
 bquitest =  executable('bquitest',
   'test/collectiontest.cpp',
   'test/databindtest.cpp',
+  'test/layouttest.cpp',
   'test/shapetest.cpp',
   'test/widgetinstancemodifiertest.cpp',
   'test/widgettest.cpp',
--- a/src/bqui/test/layouttest.cpp
+++ b/src/bqui/test/layouttest.cpp
@@ -1,0 +1,301 @@
+#include <bqui/modifier/instancemodifier.h>
+#include <bqui/modifier/setsizehint.h>
+#include <bqui/modifier/widgetmodifier.h>
+
+#include <bqui/widget/box.h>
+#include <bqui/widget/hbox.h>
+#include <bqui/widget/vbox.h>
+#include <bqui/widget/widget.h>
+
+#include <bqui/buildparams.h>
+#include <bqui/inputarea.h>
+#include <bqui/simplesizehint.h>
+#include <bqui/sizehint.h>
+
+#include <bq/signal/signal.h>
+#include <bq/signal/signalcontext.h>
+
+#include <avg/obb.h>
+#include <avg/transform.h>
+#include <avg/vector.h>
+
+#include <btl/uniqueid.h>
+
+#include <gtest/gtest.h>
+
+#include <string>
+#include <utility>
+#include <vector>
+
+using namespace bqui;
+using namespace bqui::widget;
+
+namespace
+{
+
+/**
+ * @brief Position and size a layout gave to one child.
+ *
+ * The position is in the coordinates of the layout that was realised, with
+ * the origin at its bottom-left corner.
+ */
+struct Geometry
+{
+    avg::Vector2f position;
+    avg::Vector2f size;
+};
+
+/**
+ * @brief Creates probe widgets and reads back the geometry they were given.
+ *
+ * A probe is an empty widget with a fixed size hint that tags its realised
+ * Instance with an InputArea carrying a unique id. Input areas are the one
+ * piece of geometry that travels intact from a leaf to the root Instance:
+ * every enclosing transform is accumulated into the area's own transform
+ * while its obb stays in the leaf's local coordinates. Realising the layout
+ * and looking up the areas by id therefore recovers both the size each child
+ * was allocated and where it was placed, without rendering anything.
+ */
+class ProbeSet
+{
+public:
+    AnyWidget add(SizeHintResult width, SizeHintResult height);
+
+    std::vector<Geometry> realise(AnyWidget widget, avg::Vector2f size) const;
+
+private:
+    std::vector<btl::UniqueId> ids_;
+};
+
+AnyWidget tagGeometry(btl::UniqueId id)
+{
+    return makeWidget()
+        | modifier::makeWidgetModifier(modifier::makeInstanceModifier(
+                    [id](Instance instance)
+                    {
+                        auto areas = instance.getInputAreas();
+                        areas.push_back(makeInputArea(id, instance.getObb()));
+
+                        return std::move(instance)
+                            .setInputAreas(std::move(areas));
+                    }))
+        ;
+}
+
+AnyWidget ProbeSet::add(SizeHintResult width, SizeHintResult height)
+{
+    btl::UniqueId id = btl::makeUniqueId();
+    ids_.push_back(id);
+
+    return tagGeometry(id)
+        | modifier::setSizeHint(simpleSizeHint(width, height))
+        ;
+}
+
+std::vector<Geometry> ProbeSet::realise(AnyWidget widget,
+        avg::Vector2f size) const
+{
+    auto instanceSignal = std::move(widget)(BuildParams())(
+            bq::signal::constant(size))
+        .getInstance();
+
+    // The whole tree is evaluated through this one context. A signal evaluated
+    // in a context of its own would be a different, parallel instantiation of
+    // the same graph (bq::signal::SignalContext).
+    auto context = bq::signal::makeSignalContext(std::move(instanceSignal));
+    Instance const& instance = context.evaluate<0>().get<0>();
+
+    std::vector<Geometry> result;
+    result.reserve(ids_.size());
+
+    for (auto const& id : ids_)
+    {
+        InputArea const* found = nullptr;
+        for (auto const& area : instance.getInputAreas())
+            if (area.getId() == id)
+                found = &area;
+
+        if (!found)
+        {
+            ADD_FAILURE() << "probe " << id << " was not realised";
+            result.push_back(Geometry{});
+            continue;
+        }
+
+        result.push_back(Geometry{
+                found->getTransform().getTranslation(),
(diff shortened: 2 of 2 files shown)


samienne:
Correction to the second `getSizes` fragility noted in this PR — the direction is wrong, found while folding these notes into the ArraySignal design doc (#100).

**Claim 1 (divide-by-zero) is confirmed, and is sharper than stated.** The denominator `combined[i] - combined[i-1]` is zero for any fixed hint (`{{30,30,30}}` gives `combined = {60,60,60}` for two children). At natural size that is `0/0 = NaN`; oversized it is `+inf`. Both land on `1.0f` only because `std::min(a,b)` is `(b<a)?b:a`, so `std::min(1.0f, NaN)` returns `1.0f`. Reversing the operands to `std::min(m, 1.0f)` returns `NaN`, and `std::max(0.0f, NaN)` then returns `0.0f` — **every fixed-size child collapses to zero width**. So it is a verifiable failure mode, not just a smell.

**Claim 2 (overflow) is backwards.** A 300k-case brute-force sweep over monotonic hints found **zero** overflow cases: the multiplier clamp holds the total to the aggregate maximum, so an oversized box *under-fills* and leaves slack — the opposite of placing children past its own bounds.

The underlying concern is still real, but the bound comes from **hint monotonicity**, not from a clamp. With a non-monotonic hint it is reachable: `{{100, 0, 100}}` gives `multiplier = {1, 0, 1}` and a 200-wide child in a 200-wide box, so two of them total 400. Nothing rejects such a hint today.

What a rewrite needs to take from this: do not assume `sum(getSizes(size, hints)) == size`, and do not assume the output is clamped against `size` for arbitrary hints.

samienne:
Second correction, and this one overturns my earlier comment above.

I claimed that reversing the operands to `std::min(m, 1.0f)` would make `std::max(0.0f, NaN)` return `0.0f` and collapse every fixed-size child to zero width. That was derived by simulating the arithmetic. **It is wrong.**

PR #103 tested it empirically — applied the reversed-operand mutation, rebuilt, ran the suite — and all 24 layout tests stayed green. The reason: a zero *combined* interval is a sum of non-negative per-child intervals, so when the denominator is zero every child's interval at that index is also zero. The multiplier is multiplying zero, so whatever garbage it holds is irrelevant.

So the accurate statement is: **the divide-by-zero is real, but it is not observable for monotonic hints, and the `std::min` argument order is not load-bearing.** It remains worth knowing that the NaN exists, but it is not the rewrite hazard I described.

The other half stands and is now confirmed by an actual test: `getSizes` never clamps its output against `size`, and a non-monotonic hint reaches overflow — `{{100, 0, 100}}` gives two 200-wide children in a 200-wide box. Nothing validates hints today. #103 pins that as current behaviour in `mapObbsOverflowsOnNonMonotonicHints`, so whichever fix is chosen (reject / sort / clamp) visibly changes a test.

Apologies for the churn — the difference was simulation versus running it.

PR #104, opened 2026-07-19
btl: stop merge/whenAll completing from inside init()

samienne:
Two defects in the same pair of combinators, both arising from `init()` walking its input container while completion can already be running.

## 1. The crash

`async.perf` aborts intermittently under a **debug clang-cl** build on Windows with

```
...\include\vector(78) : Assertion failed: can't increment invalidated vector iterator
```

Measured on a 16-core Windows box: **0/200** runs on an idle machine, **2/40** full-suite runs with the machine loaded. It never reproduces under MSVC `cl` debug, nor under a debugoptimized/ASan build.

`merge()` registers a completion callback on each input while iterating its input vector:

```cpp
btl::forEach(futures_, [&](auto&& future) { future.addCallback_(...); });
```

`FutureControl::addCallback` invokes the callback **inline** when the input is already ready. The outstanding-input counter therefore reaches zero on the *last* registration, so `reportFutureReady()` runs re-entrantly from inside the loop body — and it ends with `futures_.clear()`. Control returns into the range-for, which increments an iterator into the vector it just cleared. The `clear()` is the only thing that invalidates: the result-building loop above it reads through `Future::get() &&`, which never touches the `Future`'s `control_` and leaves the vector's storage alone.

This is same-thread re-entrancy, not a cross-thread data race. Instrumenting `MergedFuture` to report whether completion happened while `init()` was still walking showed `sameThread=1` in every observed instance, including the one caught in a real `async.perf` crash window. Whether the window is hit at all depends on scheduling, which is why it needs machine load and why it never showed up on an idle run.

The counter can only reach zero at the *last* registration, on either combinator: reaching zero requires every input to have completed **and** every callback to have been registered, and registration is what drives the decrements.

**Fix:** give the counter an extra count owned by `init()`, released only once the walk is finished. Completion cannot run while the container is being iterated, because the counter cannot fall below one until `init()` releases its count.

Invariant, with N registered inputs: total decrements = (number of inputs reporting success) + 1. `reportFailure` never decrements and the `failed_` early-return withholds more, so zero is reachable only when every input succeeded **and** `init()` has finished — exactly once. `assert(count >= 1)` remains unreachable.

Side effect: `merge()` of an empty vector now completes instead of never completing at all. Previously `count_` started at zero and no callback ever existed to drive it, so the future never resolved. Pinned by `async.mergeEmpty`.

## 2. Inputs retained when a merge fails

`reportFailure` guarded its `futures_.clear()` behind `initialized_`, and `init()` did its own failure clear *before* storing `initialized_`. A failure landing between init's `failed_` check and the `initialized_.store` found `initialized_` false and declined to clear, while init had already passed its own check. Nothing cleared `futures_`, so every input future and the values it holds stayed alive for as long as the merged future did.

Verified rather than argued: widening that window with a temporary sleep between the two statements and failing an input from another thread leaves a retained input value at `use_count() == 2`; with the rendezvous below, the same experiment gives `1`. (The widening was scaffolding and is not in the diff.)

**Fix:** replace both `initialized_`-guarded clears with a single arrival rendezvous, with all three parties arriving:

```cpp
if (clearArrivals_.fetch_add(1, std::memory_order_acq_rel) == 1)
    futures_.clear();
```

`fetch_add` returns the previous value, so the second arrival clears. Exactly two of the three sites ever arrive: `init()` always, and precisely one of success/failure, since a failed input calls `reportFailure` instead of `reportFutureReady` and so the counter never reaches one. Init's arrival is what marks the walk as over, so the clearing arrival can never be mid-walk. There is no read-then-act window left, and `initialized_` becomes dead and is removed from both files.

Both sites arrive **before** publishing the result, so a ready combined future implies its inputs are already released. Otherwise the guarantee would be weaker than it sounds: a thread woken by `waitForResult`, or a continuation run inline from `set()`, could observe readiness with the inputs still alive. `whenAll`'s success path builds the combined tuple first so it can drop the inputs before calling `setValue`.

Including the **success** site in the rendezvous is robustness and uniformity, **not** a live bug fix — the success path was already safe under the extra-count scheme above, since workers decrementing during the walk bottom out at one and only init's trailing decrement can complete. Folding it in makes the invariant local to `arriveAndClear()` instead of resting on a counting argument that a later change to `count_` could silently invalidate.

## `whenall.h` status

`whenAll` has both defects in the same shape — `values_.reset()` destroys the tuple that the `tuple_foreach` walk still refers to, and the identical `initialized_` window — and **both are fixed here**, identically. `tuple_foreach` expands `std::get<S>(data)` per element, so any tuple element after the last future is visited after a reset.

Stating the asymmetry plainly, since it affects how much this PR proves:

- The `merge` crash is **observable**: a vector iterator is invalidated and the MSVC debug iterators abort on the next increment. `async.mergeInputsReadyBeforeInit` reproduces it 100% of the time against the unpatched header.
- The `whenAll` **crash** variant is **not observable on any configuration tested** — no container iterator for the debug STL to check, the destroyed tuple lives in storage that is destroyed but not freed so ASan sees nothing, and the trailing-element visit forms a reference without reading through it. `async.whenAllInputsReadyBeforeInit` passes against the unpatched header, so it pins behaviour but is not a reproducer.
- The `whenAll` **retention** variant *is* observable, and `async.whenAllReleasesInputsOnFailure` covers it.

## Tests, and what they actually assert

- `async.mergeInputsReadyBeforeInit` — every input ready before registration. Aborts 100% against the unpatched header; passes with the fix. Verified by A/B swapping only the two headers.
- `async.mergeEmpty` — pins that an empty merge completes. Cannot be an A/B test: pre-fix this *hangs* rather than fails.
- `async.mergeReleasesInputsOnFailure`, `async.whenAllReleasesInputsOnFailure` — assert observable lifetime (`use_count()` of a value owned by a retained input), not internals, sampled both *inside* the failure continuation and after it. Confirmed non-vacuous twice over: deleting the failure-path arrival makes both fail with `use_count() == 2`, and merely moving that arrival back to *after* `setFailure` also makes both fail, with `useCountWhenReady == 2`.
- `async.whenAllInputsReadyBeforeInit` — pins behaviour only, as above.

Two things could **not** be tested honestly, rather than being covered by something that passes vacuously:

- The exact racy interleaving of defect 2 is a two-statement window reachable only cross-thread. Pinning it needs a hook inside the header; it was verified by the widened-window experiment described above instead.
- The **success**-path clear is not observable through input values, because success moves the values out into the result and leaves a moved-from husk behind. It is pinned only behaviourally, by the existing success tests.

## Why `acq_rel` on the counter

The mutual exclusion does **not** depend on the ordering — it comes from the atomicity of the counter alone. Whichever decrement is last in the modification order of `count_` completes, and the decrement from `init()` is by construction not last unless the walk is over. That holds under any ordering.

What the ordering buys is that the *data* the winner reads is well-defined rather than accidentally correct. The interleaving it covers:

1. The init thread registers the callback for the final input, which is still pending; `addCallback` stores it and returns.
2. The init thread is preempted inside `forEach`, before finishing the walk.
3. A worker completes that input, runs the callback, and decrements. With the fix its `fetch_sub` returns 2, not 1, so it does not complete — `init()` still holds its count. Init resumes, finishes the walk, decrements, gets 1, and completes after the walk.
4. The mirror case: init's decrement lands first and returns 2; a straggling worker decrements, gets 1, and *that* worker completes, then reads and moves out of `futures_`.

In step 4 the completing thread is not the one that ran `init()`, so it must observe everything `init()` did before releasing its count. That needs init's decrement to be a **release** and the completing thread's to be an **acquire**; either thread can play either role depending on which lands last, so both are needed on the same operation — hence `acq_rel`. Read-modify-write operations on `count_` form a release sequence, so the acquiring thread still synchronizes-with the release when other decrements intervene. The same reasoning applies to the arrival counter, which carries the "the walk is over" publication.

## Verification

| config | before | after |
|---|---|---|
| clang-cl debug, full suite under load | 2/40 abort | 0 aborts / 270 |
| clang-cl debug, `merge` regression test | 100% abort | pass |
| clang-cl debug, retention experiment (widened window) | `use_count` 2 | `use_count` 1 |
| MSVC cl debug | does not detect the crash | pass |
| clang-cl debugoptimized + ASan | clean | clean |
| clang-cl debug, `whenAllFail`+`futureFail` under load | 3/400 fail (master) | 0/1600 |

`meson test` is green in all three build configurations.

## 3. A pre-existing test flake (third commit, droppable)

Surfaced by the soaks for the work above, but **present on master and unrelated to it**: `async.whenAllFail` fails its `EXPECT_TRUE(failCalled.load())` at roughly 1 run in 400 under load on Windows/clang-cl. Measured on **pristine `origin/master`: 3/400**; on this branch **2/400** and **1/400** — indistinguishable, so not introduced here.

The tests asserted that their `onFailure` callback had run as soon as `get()` returned. Nothing promises that. `FutureControl::set()` publishes `ready_` at `futurecontrol.h:215` and only invokes the callback at `:224`; `waitForResult`'s **slow** path is fine, since it stashes `callback_`, installs a notifier and runs the old callback at `:69-70` before returning — but the **fast** path at `:44-45` (`if (ready()) return;`) returns without touching `callback_` at all. Hence a *late* callback rather than a lost one, and only when the value is already published by the time the waiter looks. A probe that keeps waiting past the assertion point found **1 late and 0 lost over 18,000,000 iterations**.

Fixed on the **test** side. Guaranteeing the ordering in the library would need a second synchronization point on the `waitForResult` fast path that every future in the system pays for, on behalf of these tests. The tests now wait on the callback itself, through a promise the callback fulfils. `async.futureFail` had the same pattern plus a plain `bool` written by a worker and read by the test thread, and is fixed the same way.

Checked before choosing the test side: **nothing in the codebase relies on "value observed implies continuations have run"**, so the guarantee is neither real nor depended upon. Searched every `btl::future` consumer outside `btl` — `ase`'s fence futures in `glxplatform`/`wglplatform` (`wait()` for frame throttling, then `pop()`, with no continuation on the waited future), `ase/commandbuffer.cpp`'s fire-and-forget `pushFence().then(cb).detach()`, and `bq/test/signaltest.cpp` (whose `.then()` runs inline because the futures are already ready, making the later `wait()` a no-op). `bqui/src/app.cpp` and `bq/signal/{signal,shared}.h` include the headers without using them; `avg` and `testapp1` have no usage. The `merge` hits in `bqui` are `bq::signal::merge`, an unrelated function.

Soak: **0 failures in 1600 runs** of `async.whenAllFail` + `async.futureFail` under load, against a pre-fix baseline of **3/400** on master. At that baseline 1600 runs would be expected to produce roughly 12 failures, so observing none is not luck (P(0) ≈ 6e-6).

**This is still a soak result, not a proof.** The rewritten tests are not deterministic the way the manual-future tests are — they run real threadpool work, and the fix removes a wrong assertion rather than removing the concurrency. The claim is that the assertion no longer depends on unsynchronized ordering, and the soak is consistent with that.

## Notes

- MSVC `cl` does not flag the crash at `_ITERATOR_DEBUG_LEVEL=2` even though clang-cl does; a four-line standalone repro of "clear a vector inside a range-for" survives under `cl` and aborts under `clang-cl` with identical flags. The clang-cl report is correct — the code is UB either way.
- ASan reports nothing because nothing is *read* after the invalidation: the cleared vector keeps its buffer, and the invalidated iterator increments straight onto the cached `end()`. The impact is confined to iterator-checked builds; it is still UB, and it aborts a first-class CI toolchain.
- This is **not** the long-standing macOS/ARM64 `whenAll` flakiness, and this PR does not claim to fix it.
- The pre-existing `async.whenAllFail` flake is handled in the third commit — see the section below. It is deliberately separable from the two fixes above.
- Unrelated defect spotted while reading `threadpool.h`, **not** fixed here: `TimedQueue::push()` does a plain `push_back` with no `std::push_heap`, while `pop()` calls `std::pop_heap` and `hasTaskReady()`/`getNextEventTime()` read `front()` as if it were the earliest. With two or more outstanding timed tasks they can fire out of order. Only `btl::delayed` uses this path and never with more than one task at a time, so nothing currently exercises it.

### DIFF
--- a/src/btl/include/btl/future/merge.h
+++ b/src/btl/include/btl/future/merge.h
@@ -49,10 +49,7 @@ namespace btl::future
 
             });
 
-            if (failed_.load(std::memory_order_acquire))
-                futures_.clear();
-
-            initialized_.store(true, std::memory_order_release);
+            arriveAndClear();
 
             reportFutureReady();
         }
@@ -63,10 +60,9 @@ namespace btl::future
             bool hadNotFailed = failed_.compare_exchange_strong(expected, true);
             if (hadNotFailed)
             {
-                this->setFailure(std::move(err));
+                arriveAndClear();
 
-                if (initialized_.load(std::memory_order_acquire))
-                    futures_.clear();
+                this->setFailure(std::move(err));
             }
         }
 
@@ -82,16 +78,29 @@ namespace btl::future
                 result.reserve(futures_.size());
                 for (auto&& future : futures_)
                     result.push_back(std::move(future).get());
-                this->setValue(std::move(result));
 
-                futures_.clear();
+                arriveAndClear();
+
+                this->setValue(std::move(result));
             }
         }
 
     private:
+        // Arrivals come from init(), the success path and the failure path,
+        // and exactly two of them ever happen: a failed input reports failure
+        // instead of readiness, so count_ never reaches one. init() arrives
+        // once its walk is over, so the second arrival, the one that clears,
+        // never runs mid-walk. Both callers arrive before publishing the
+        // result, so a ready merged future has already dropped its inputs.
+        void arriveAndClear()
+        {
+            if (clearArrivals_.fetch_add(1, std::memory_order_acq_rel) == 1)
+                futures_.clear();
+        }
+
         std::atomic_int count_;
         std::atomic_bool failed_ = false;
-        std::atomic_bool initialized_ = false;
+        std::atomic_int clearArrivals_ = 0;
         std::vector<Future<T>> futures_;
     };
 
--- a/src/btl/include/btl/future/merge.h
+++ b/src/btl/include/btl/future/merge.h
@@ -13,8 +13,11 @@ namespace btl::future
         public FutureControl<std::vector<T>>
     {
     public:
+        // count_ starts one above the input count; init() owns the extra count
+        // and releases it only after it has stopped walking futures_, so
+        // completion never destroys futures_ from under that walk.
         MergedFuture(std::vector<Future<T>> futures) :
-            count_(static_cast<int>(futures.size())),
+            count_(static_cast<int>(futures.size()) + 1),
             futures_(std::move(futures))
         {
         }
@@ -50,6 +53,8 @@ namespace btl::future
                 futures_.clear();
 
             initialized_.store(true, std::memory_order_release);
+
+            reportFutureReady();
         }
 
         void reportFailure(std::exception_ptr err)
@@ -67,7 +72,7 @@ namespace btl::future
 
         void reportFutureReady()
         {
-            auto count = count_.fetch_sub(1, std::memory_order_acquire);
+            auto count = count_.fetch_sub(1, std::memory_order_acq_rel);
 
             assert(count >= 1);
 
--- a/src/btl/include/btl/future/whenall.h
+++ b/src/btl/include/btl/future/whenall.h
@@ -51,8 +51,11 @@ namespace btl::future
                     >
     {
     public:
+        // count_ starts one above the input count; init() owns the extra count
+        // and releases it only after it has stopped walking values_, so
+        // completion never destroys values_ from under that walk.
         WhenAll(std::tuple<Ts...> values) :
-            count_(((IsFuture<std::decay_t<Ts>>::value ? 1 : 0) + ... + 0)),
+            count_(((IsFuture<std::decay_t<Ts>>::value ? 1 : 0) + ... + 0) + 1),
             values_(std::move(values))
         {
         }
@@ -86,6 +89,8 @@ namespace btl::future
                 values_.reset();
 
             initialized_.store(true, std::memory_order_release);
+
+            reportValueReady();
         }
 
         void reportFailure(std::exception_ptr err)
@@ -103,7 +108,7 @@ namespace btl::future
 
         void reportValueReady()
         {
-            auto count = count_.fetch_sub(1);
+            auto count = count_.fetch_sub(1, std::memory_order_acq_rel);
 
             assert(count >= 1);
 
--- a/src/btl/include/btl/future/whenall.h
+++ b/src/btl/include/btl/future/whenall.h
@@ -85,10 +85,7 @@ namespace btl::future
 
             });
 
-            if (failed_.load(std::memory_order_acquire))
-                values_.reset();
-
-            initialized_.store(true, std::memory_order_release);
+            arriveAndReset();
 
             reportValueReady();
         }
@@ -99,10 +96,9 @@ namespace btl::future
             bool hadNotFailed = failed_.compare_exchange_strong(expected, true);
             if (hadNotFailed)
             {
-                this->setFailure(std::move(err));
+                arriveAndReset();
 
-                if (initialized_.load(std::memory_order_acquire))
-                    values_.reset();
+                this->setFailure(std::move(err));
             }
         }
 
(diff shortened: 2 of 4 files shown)


src/btl/include/btl/future/merge.h
```diff
@@ -77,16 +78,26 @@ namespace btl::future
...
-                futures_.clear();
+                arriveAndClear();
+
+                this->setValue(std::move(result));
             }
         }
 
     private:
+        // The second of the two arrivals clears, so the inputs are gone before
+        // the merged future is made ready. See src/btl/AGENTS.md for why
```
samienne: Let's preferably not refer to the documentation as it is quite easy to go stale. Especially let's not refer to the agent documentation. Either give a brief explanation here or just leave it out.

src/btl/include/btl/future/whenall.h
```diff
@@ -94,40 +96,50 @@ namespace btl::future
...
-                values_.reset();
+                arriveAndReset();
+
+                this->setValue(std::move(value));
             }
         }
 
     private:
+        // The second of the two arrivals resets, so the inputs are gone before
+        // the combined future is made ready. See src/btl/AGENTS.md for why
```
samienne: Also here, remove the reference to agent docs.

PR #110, opened 2026-07-21
Revise the ArraySignal design: state belongs in the SignalContext

samienne:
Supersedes the design half of #100. The implementation on that PR is a spike and should not be extended.

## Why

`OncePerId` keeps its per-identity table in a `shared_ptr` captured by a `map` lambda (`arraysignal.h:185-191`), so the table lives in the signal **description**, not in the `DataContext`. `bq`'s rule is the opposite — signals are stateless graph descriptions and state lives in the context (`src/bq/AGENTS.md`), which is what `SharedControl` implements via `context.findData` keyed by a `UniqueId` (`sharedcontrol.h:147-166`).

Two `SignalContext`s over one `ArraySignal` therefore share one table and evict each other's entries. Those entries are what the delegate built — widget state — so this is a correctness defect, not the bounded churn the first draft recorded. Every `.share()`-placement puzzle in that draft was a symptom: `.share()` was being used to stop the table being instantiated twice, which is the job `DataContext` keying does for free.

## What changes

**Five operators disappear.** `KeyedArraySignal`, `extract`, `mapValues`, `mapKeyed` and `values` existed only because there was no node that could hold state, so anything collective had to leave the domain and come back. `mapValues` was a plain `bq::map` on the flattened vector all along. What remains is construction, `forEach`, `concat`, `map`, `scatter`, `join`.

**The framing "an analog of `AnySignal<vector<T>>`" is withdrawn** — it is what invited the round trip. `ArraySignal<U>` is a reactive collection of stable identities each carrying a `U` built once; it collapses to `AnySignal<vector<X>>` only when `U` is an `AnySignal<X>`, and that is the only exit. Checked against the layout pipeline rather than assumed: both fan-ins land in that form.

**Ids are minted per context at initialisation**, so `concat` needs no re-namespacing and the caller's key contract weakens to uniqueness *within one `forEach`*.

**No bespoke signal type is needed.** The per-element `AnySignal<T>` is `sharedKeyedSource.map(pick(key))`, where the shared source supplies per-context state through `SharedControl`. `pick` latches its last value so a departed key has a defined value rather than relying on an ordering argument.

**`AnySignal<vector<T>>` stops being a constructor.** In the spike it keyed by position (`arraysignal.h:450-457`), which is what made one sibling insert take the build count 1 → 4. Use `forEach` and say what the key is.

**The shared form is structurally required**, not an optimisation: every container branches (hints and instances read the same builders), so an unshared description would call the delegate twice.

## Also recorded

- `DataContext::initializeData`'s assert (`datacontext.h:70`) is now a **prerequisite** fix, not a nicety — `data_` holds `weak_ptr`s, the assert requires absence, and shrink-then-grow is a legal sequence that fires it.
- The two `bq` `join` bugs the spike found are real and independent; they should land on their own.
- The layout audit, the rationale for a signal-taking delegate, and the `getSizes` hazards are unchanged.

Doc only — no code.

### DIFF
--- a/docs/design/arraysignal.md
+++ b/docs/design/arraysignal.md
@@ -514,6 +514,12 @@ same relationship `Signal` has to a `.share()`d signal: `SharedArraySignal`
 carries a `UniqueId`, and every consumer instantiating it in one context finds
 the same table via `findData`.
 
+**Sharing within one context is a supported, existing mechanism, not something
+this design has to invent.** `SharedSignal` is the reference implementation —
+follow it. Sharing *between* contexts is the thing that must not happen, and the
+two are easy to conflate: the first implementation's failure was that it had no
+way to tell them apart, because its table was not keyed by context at all.
+
 **Implement only the shared form.** The consuming variant would have no user on
 day one, since every container branches. Its `map` therefore takes `U const&`
 rather than consuming. The consuming variant can be added later if a linear
@@ -1208,7 +1214,9 @@ consumer-side reconcile. See *Open questions*.
 
 ## Open questions
 
-Points the design does not settle. Flagged rather than guessed.
+Points the design does not settle. Flagged rather than guessed. Sharing within a
+single context is *not* one of them — it is a supported mechanism with an
+existing reference implementation; see *Sharing: `SharedArraySignal`*.
 
 - **Does any identity surface at the exit?** Unchanged from the first draft, and
   still the one open question that changes code outside this design. *Identity is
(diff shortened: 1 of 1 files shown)


samienne:
**Do not merge on green.** This is part of the `ArraySignal` stack, and the whole stack is gated on one thing: the design being verified by a *full* implementation and then by the reworked layouts. Until `layout()` is unified and `dynamicBox` is gone, a green build here only says this stage compiles and passes its own tests — not that the design holds.

Stack order (each based on the one above):

1. #110 — design revision (doc)
2. #111 — steps 0-1: `DataContext` assert, the pick construction
3. _next_ — step 2: the array node (`forEach`, `concat`, `map`, `join`)
4. _then_ — step 3: `scatter`
5. _then_ — steps 4-5: unify `layout()`, delete `dynamicBox` — **this is the gate**

Kept as separate PRs so each stage stays reviewable; may be squashed or re-split before merging.

PR #111, opened 2026-07-21
Prove the ArraySignal pick construction

samienne:
Steps 0 and 1 of the implementation order in `docs/design/arraysignal.md`, stacked
on #110 (which carries the revised design). Nothing here is public API yet, and
nothing depends on it: `forEach`, the array node, `concat`, `map`, `scatter` and
`join` are step 2+ and land separately.

## Step 0 — `DataContext::initializeData`

The context holds weak references, so an entry whose data has been released
lingers in the map. A key that goes away, is dropped by every consumer, and then
comes back has to initialize over that stale entry — a legal sequence that fired
the debug assert. The assert now only rejects an id whose data is still alive.

`src/bq/test/datacontexttest.cpp` exercises the expire-then-reinitialise sequence
directly, and `pick.canBeInitializedAgainAfterItsStateExpired` covers it through
a real signal graph. Both abort on the previous assert.

## Step 1 — the pick construction

Three function templates in `bq::signal::detail`, in `detail/pick.h`. **No custom
signal node and no per-context state** — the whole construction is `map` and
`share`:

- `shareKeyed(source, keyFn)` maps a signal of a vector once into a signal of a
  key-to-element map and shares it, so following an element is a lookup rather
  than a scan.
- `pick(shared, key)` is a plain `map` yielding `AnySignal<std::optional<T>>`,
  empty for any update in which the key is absent.
- `requirePresent(sig, description)` is a second plain `map` that drops the
  optional and throws, for a consumer that needs the element to be there — the
  `forEach` delegate will.

### Why an optional rather than a latch

The design doc specified a **latch** (hold the last value seen when the key
leaves). That is withdrawn, and the doc is corrected here. A key's presence
cannot be encoded in the type system and cannot be guaranteed by construction, so
its absence has to surface at runtime; latching turns that into a silently stale
value, which hides the bug that caused it. It also did not buy what it promised —
it removed the ordering argument from `evaluate` only by moving it to
`initialize`, where a context holding no live instantiation had nothing to latch
and had to fail anyway.

The obligation stands and is now enforced loudly instead: build a pick when its
key appears, and drop it no later than the update that observes the key leaving.

## What the tests establish

`src/bq/test/picktest.cpp`, 10 tests, all green:

1. A pick follows its key across inserts, removals and moves of other elements,
   with the picked element changing at each step so a pick that stopped
   following would be caught. A pure reorder is invisible to it.
2. An absent key yields `nullopt` — not a throw, not a stale value — and a key
   that leaves and returns is picked up again. `requirePresent` throws both when
   the key leaves and when it was never there, and a duplicate key throws too;
   each is matched on its message.
3. Two `SignalContext`s over one description are wholly independent: each
   evaluates the source for itself, and driving one leaves the other exactly
   where it was, in both directions.
4. One description instantiated twice into **one** context finds the shared
   source's state already there and evaluates it once per pass, not once per
   consumer — pinned both directly against a `DataContext` and through a
   `SignalContext`.

## Findings recorded rather than worked around

From the two clean-context review rounds; all are in the design doc.

- **A shared node copies its value into every consumer on every changed frame**
  (`sharedcontrol.h:103`, `:199`). Sharing the map by value would cost one full
  copy per consumer per update — the same O(n²) keying exists to remove — so the
  shared signal carries a `shared_ptr` to an immutable map. The constraint
  belongs to `Shared`; a by-reference evaluate path would lift it.
- **A pick reports a change whenever any element changes**, not only its own,
  since that is what the shared source reports and the element type carries no
  comparison requirement. `tryCheck()` at the point of use is bq's existing
  answer, so this is documented and tested rather than worked around.
- **`b_ndebug` is off project-wide**, so `assert` is live in every build type. A
  debug-assert / release-throw pair does not exist here — the assert would always
  win and the throw would be unreachable — so hard errors are a throw alone. That
  is why a duplicate key throws rather than asserting.
- **A throw ends the pass, not just the node.** With a `check` or `share`
  downstream the exception surfaces from `SignalContext::update()` with the
  context part-advanced. Nothing is corrupted and the next frame recovers, but
  these are discard-the-context errors, not recoverable ones.
- `shareKeyed` invokes `keyFn` as const, so a stateful key function no longer
  compiles: its state would live in the description, shared by every context.

Also noted for follow-up, not fixed: `DataContext` never prunes expired entries,
which becomes load-bearing once ids are minted at churn rate.

`src/bq/AGENTS.md` gains a threading section recording why per-context signal
state needs no lock — a `DataContext` belongs to one `SignalContext` and is
advanced by a single synchronous pass, so concurrency in bq is between contexts.
That is independent of this construction and is the durable home for the rule.

Built and tested green under clang-cl 18 (Debug and Release) and MSVC 17 (Debug).

### DIFF
--- a/docs/design/arraysignal.md
+++ b/docs/design/arraysignal.md
@@ -495,6 +495,22 @@ absent. One copy, per-context, entirely legal, and it turns the failure mode fro
 undefined into briefly stale. The invariant then becomes something the tests
 assert rather than something the design leans on.
 
+**The construction is built and proven.** `bq::signal::detail::Pick` and its two
+helpers live in `src/bq/include/bq/signal/detail/pick.h`, with
+`src/bq/test/picktest.cpp` covering the four cases step 1 called for. Three
+results settle points this document had left open or inferred:
+
+- Two `SignalContext`s over one `pick` description are wholly independent —
+  separate latches, and each evaluates the shared source for itself.
+- One description instantiated twice into **one** context shares its latch and
+  evaluates the shared source once per pass, exactly as inferred from
+  `SharedControl`. An instantiation made after the key has already left the
+  source finds the earlier latch rather than failing for want of a value, so the
+  sharing is of the state itself and not a coincidence of two copies having seen
+  the same values.
+- `pick` reports **no change** while its key is absent, so a latched element is
+  inert rather than merely stale.
+
 **`scatter` is the same node.** Its aggregate is positional rather than keyed, but
 the scatter node knows the current key order, so it zips aggregate-with-keys into
 an `AnySignal<std::map<ArrayId, W>>`, shares that, and hands each identity the
@@ -1218,11 +1234,6 @@ Points the design does not settle. Flagged rather than guessed.
   `AnySignal<vector<pair<size_t, Window>>>` and reconciling by hand. The two live
   consumers are PR #99's window list and `dataBind`
   (`src/bqui/include/bqui/databind.h:22`). Decide it when PR #99 is reworked.
-- **Does a description instantiated twice into the *same* context share
-  correctly?** `SharedArraySignal` should get this for free — same `UniqueId`,
-  same `findData` hit, exactly as `SharedControl` does — but this is inference
-  from reading `sharedcontrol.h`, not something the spike tested. It is the first
-  thing step 1 must prove.
 - **The reactive subtree `AnySignal<ArraySignal<T>>`.** Still unimplemented and
   still unreconciled: the identity algebra says only construction and `forEach`
   mint identity, but this constructor hands out fresh ids on every swap. Either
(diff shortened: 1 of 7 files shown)


samienne:
**Do not merge on green.** This is part of the `ArraySignal` stack, and the whole stack is gated on one thing: the design being verified by a *full* implementation and then by the reworked layouts. Until `layout()` is unified and `dynamicBox` is gone, a green build here only says this stage compiles and passes its own tests — not that the design holds.

Stack order (each based on the one above):

1. #110 — design revision (doc)
2. #111 — steps 0-1: `DataContext` assert, the pick construction
3. _next_ — step 2: the array node (`forEach`, `concat`, `map`, `join`)
4. _then_ — step 3: `scatter`
5. _then_ — steps 4-5: unify `layout()`, delete `dynamicBox` — **this is the gate**

Kept as separate PRs so each stage stays reviewable; may be squashed or re-split before merging.

PR #112, opened 2026-07-21
Add the array node: ArraySignal, forEach, concat, map and join

samienne:
Step 2 of `docs/design/arraysignal.md`, stacked on #111 (steps 0-1). Draft: it
will not merge until the layouts have been reworked on top of it.

### The two `join` fixes, first and on their own

Both are pre-existing and independent of this design; the superseded spike on
#100 found them.

- `Signal::join()` instantiated `Join<TStorage>`, but `TStorage` is `void` for an
  `AnySignal`; the storage type is named by `StorageType`. The free `join()`
  deduced the same `void` from a derived-to-base match.
- `detail::toSignal` was an overload pair, and an `AnySignal` is derived from
  `Signal`, so the forwarding overload was the better match and a type-erased
  inner signal was wrapped in a `constant` instead of joined. It is now one
  function with an `if constexpr`.

### The node

All durable state is in the `SignalContext`. One node, `detail::ArrayOnce`,
holds the key-to-id map and the built values in its `DataType`; the array holds
that node behind a `.share()`, so `SharedControl` keys the data by a `UniqueId`
carried in the description. `forEach` and `map` are that node with different
key, build and id functions.

`forEach` hands its delegate the pick construction from step 1, and honours what
`requirePresent` documents: a pick is built when its key appears and dropped in
the update that observes it leave.

### Acceptance tests

The three the design names, all in `src/bq/test/arraysignaltest.cpp`:

- shrink then grow;
- remove while a sibling still reads — a fold inside a surviving element's `U`
  is untouched by a neighbour's insertion or removal;
- two contexts, diverging membership.

The sibling test is load-bearing: dropping `join`'s per-identity `.share()`
reproduces the spike's failure exactly.

### Where the design did not survive contact

Recorded under *Corrections from the implementation* in the design doc, with the
surrounding prose amended:

- there is no separate `SharedArraySignal` — sharing is intrinsic, so copies of a
  description share within one context and only a second context is independent;
- `join` must share each element's signal once per identity, or a membership
  change restarts every survivor's per-instantiation state;
- a value change at fixed identity is a **new** identity, not a new value at the
  old one, otherwise `map` silently ignores it — which is `dynamicBox`'s bug;
- the doc's two accounts of `map` were incompatible; the once-per-identity one is
  what is built;
- no per-array id space is needed.

Out of scope and untouched: `scatter` (step 3), the `AnySignal<ArraySignal<T>>`
reactive subtree, and everything in `bqui`.

Verified with MSVC and clang-cl, Debug, whole project: builds clean and all four
test suites pass.

### DIFF
--- a/src/bq/include/bq/signal/arraysignal.h
+++ b/src/bq/include/bq/signal/arraysignal.h
@@ -6,7 +6,6 @@
 #include "constant.h"
 #include "datacontext.h"
 #include "frameinfo.h"
-#include "join.h"
 #include "signal.h"
 #include "signalresult.h"
 #include "signaltraits.h"
@@ -111,15 +110,13 @@ namespace bq::signal
 
         /** @brief Whether `U` is accepted as a single constant item of `T`.
          *
-         * A signal and an array of their own have their own constructors, so an
-         * unconstrained forwarding constructor would swallow both — and it
-         * would be a better match than the copy constructor for a non-const
-         * array lvalue besides.
+         * An array has its own constructors, so an unconstrained forwarding
+         * constructor would swallow it — and it would be a better match than
+         * the copy constructor for a non-const array lvalue besides.
          */
         template <typename U, typename T>
         constexpr bool isPlainItem =
             std::is_convertible_v<U, T>
-            && !std::is_convertible_v<U, AnySignal<T>>
             && !IsArraySignal<std::decay_t<U>>::value;
 
         /** @brief Builds one value per key and keeps it while the key is
@@ -265,41 +262,31 @@ namespace bq::signal
                         std::move(idFunc)));
         }
 
-        /** @brief One element whose value comes from a signal.
+        /** @brief One element that never changes.
          *
-         * A new value is a new identity. Nothing built this value from an
-         * identity that outlives the change, so whatever an operator built
-         * once per identity has to be built again — which is the rebuild an
-         * array of a varying single item is defined to cost. Only this element
-         * is disturbed.
+         * It has no inner signal at all. The identity is minted when the
+         * element is initialized into a context, so the description carries
+         * none and two contexts over it are unrelated.
          */
         template <typename T>
-        class ArraySingle
+        class ArrayConstant
         {
         public:
-            using Storage = typename AnySignal<T>::StorageType;
             using Elements = ArrayElements<T>;
 
             struct DataType
             {
-                typename Storage::DataType innerData;
                 Elements elements;
             };
 
-            ArraySingle(Storage sig) :
-                sig_(std::move(sig))
+            ArrayConstant(T value) :
+                value_(std::move(value))
             {
             }
 
-            DataType initialize(DataContext& context,
-                    FrameInfo const& frame) const
+            DataType initialize(DataContext&, FrameInfo const&) const
             {
-                DataType data { sig_.initialize(context, frame), {} };
-
-                setValue(data, sig_.evaluate(context, data.innerData)
-                        .template get<0>());
-
-                return data;
+                return DataType { Elements { { makeArrayId(), value_ } } };
             }
 
             SignalResult<Elements const&> evaluate(DataContext&,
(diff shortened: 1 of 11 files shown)


samienne:
Both clean-context reviews are in and applied; `6bd2eca` is the result.

**Real bugs the correctness review caught, now fixed and tested**

- `map` forwarded its source's identities, so `concat(a, a.map(f))` threw
  `duplicate key` — and that is exactly the shape of the branch a container makes
  and rejoins. Every array-building node now mints its own identities.
- A varying single item minted a new identity on every *reported* change, not
  every value change, so a no-op upstream set rebuilt the element. It is
  deduplicated with `tryCheck()` now — which, per `tryCheck`'s own trap, protects
  only equality-comparable `T`.

**Recorded rather than fixed**

`join` builds its per-identity `.share()` at update time, so it mints a
`DataContext` id per element identity per context — and `DataContext` never
prunes. Every other `initializeData` caller mints at description-construction
time, so this is the first node that grows `data_` at churn rate. It was already
noted in the design as a future direction; it is now a **prerequisite for step
4**, and the implementation order says so. Fixing it is a change to a core `bq`
data structure and belongs in its own PR.

Also documented rather than changed: `Join::update` builds the new inner data
before releasing the old, which is what lets a surviving element keep its state.
That is now load-bearing for a contract in `arraysignal.h`, so `join.h` says so —
rewriting it as a reset-then-assign would silently restart every survivor's folds
with only one test to catch it.

Style review findings applied: the `Any`-prefix carve-out for `ArraySignal` is
recorded in `docs/conventions.md`, the SFINAE is extracted to
`detail::isPlainItem`, and the id-counter claim in the design doc was wrong and
is corrected.

MSVC and clang-cl, Debug, whole project: clean, all four suites pass.

src/bq/include/bq/signal/arraysignal.h
```diff
@@ -0,0 +1,632 @@
...
+     *
+     * `keyFn` and `delegate` are invoked as const functions, and run for every
+     * item on every change and once per key respectively. Keys must be unique
+     * within this call and need not be unique anywhere else.
+     *
+     * @throws std::runtime_error if two items share a key. Thrown from an
+     *         update, this leaves the source advanced and the table not, so
+     *         the context should be discarded rather than driven again.
+     */
+    template <typename T, typename TKeyFunc, typename TDelegate>
```
samienne: TKeyFunc and TDelegate must be restricted with std::is_invocable(_r). Also Signal<TStorage, std::vector<T>> might not be able to match AnySignal<std::vector<T>> so it must be something like Signal<TStorage, std::vector<T>> to match any type of signal that provides std::vector<T>. I wonder if we allow the values to be convertible as well.

src/bq/include/bq/signal/arraysignal.h
```diff
@@ -0,0 +1,710 @@
...
+    {
+        /** @brief The identity of one element of one array.
+         *
+         * An id never reaches user code: it is minted by an array's node, in
+         * that node's per-context state, and is only ever matched within the
+         * node that minted it. It is orderable so that it can key a map;
+         * nothing else about it means anything, and in particular its order
+         * says nothing about the order of the elements.
+         */
+        class ArrayId
```
samienne, line 47: Why is this even needed, why not just use UniqueId?

samienne:
Reworked as directed, plus a second round of both clean-context reviews. Tip is
`ff4d59d`.

## The exit is now its own node

`detail::ArrayJoin` holds one inner data per identity in its per-context
`DataType`, keyed by `ArrayId`. An update carries a survivor's data across
untouched, initialises only arrivals, releases only departures. The per-identity
`.share()` is gone.

`aSurvivingElementKeepsItsStateThroughItsNeighbours` still passes, and now for
the right reason — making the node re-initialise survivors instead of carrying
them across is what breaks it (checked).

**The churn is gone.** The array machinery mints no `DataContext` entry per
identity at all now: an `ArrayId` keys only the nodes' own tables, and every
`share()` left in the implementation is built when the description is. What still
mints at churn rate is a *delegate* that builds something stateful — a widget's
`Input`s and `share()`s — which is inherent to building per item and is not
something this design can avoid. *Future directions* says exactly that instead of
claiming a prerequisite.

## The varying-item constructor is gone

Every constructor is constant. Only construction of a constant item and `forEach`
mint identity, `map` runs exactly once per identity because the value it reads
cannot change, and the doc's two accounts of `map` are one account again. Three
of the five corrections went with it; the two that survive are folded into the
prose and the section is deleted.

## What else was resting on the generic `join`

Only the per-identity share, and the correction that described it. Nothing else
had to be replaced — `ArrayOnce`, the pick construction and `concat` are
untouched by the rework. The two `join` fixes still stand on their own: they are
needed by `Signal::join` generally, the array exit simply no longer uses it.

## Second-round review findings, fixed

- **An arriving element was advanced twice.** It was initialised against the
  update's frame and then updated in it, so a fold inside it counted its first
  value twice. The existing tests missed it because the item signal was
  deduplicated with `check()`; the new test deliberately is not.
- **The exit did not reject a repeated identity.** `concat(a, a)` gave the two
  occurrences one inner data and advanced it twice a frame, so whether the
  documented duplicate error appeared depended on which operator you reached for.
- **`reconcile` was not exception-safe** — a throw from an arrival's initialise
  left the data half-drained, so a later `evaluate()` would crash rather than
  throw. Arrivals now initialise into their own storage first.

## Two things stated rather than fixed

- **Two consumers of one array do not share the elements' instantiated state**,
  only the built values — each exit gives every element its own inner data, as
  instantiating any signal twice in one context does. The delegate and every
  `map` still run once between them. The class doxygen claimed one evaluation per
  pass; corrected.
- **`ArrayJoin::observe` is not rewired by a membership change.** Nothing in the
  toolkit observes a signal today and `Join::observe` has the same shape, so it is
  documented rather than solved; solving it needs a lifetime protocol for
  per-element connections that `btl::connection` does not currently support.

MSVC and clang-cl, Debug, whole project: clean, all four suites pass.

src/bq/include/bq/signal/arraysignal.h
```diff
@@ -0,0 +1,759 @@
...
+        }
+
+        /** @brief One element that never changes.
+         *
+         * It has no inner signal at all. The identity is minted when the
+         * element is initialized into a context, so the description carries
+         * none and two contexts over it are unrelated.
+         */
+        template <typename T>
+        class ArrayConstant
```
samienne: I wonder if this is necessary and could be done with just signal::constant. I agree that it is nice for symmetry but I think whenever this is used we could just use constant. If needed this could be turned into a helper function.

samienne:
All three notes addressed in `86ddac5`.

**1. Callables constrained, source widened.**

`forEach` now takes any `Signal<TStorage, std::vector<T>>`. You were right to
want it checked rather than reasoned about: an `AnySignal<std::vector<T>>` does
bind, through the `Signal<void, ...>` base it derives from, but the *old*
signature was the broken direction — it took `AnySignal<std::vector<T>>`, which
deduces `T` only from that exact type, so a concrete signal had to be erased at
the call site first. Two tests pin both bindings
(`forEachTakesAnErasedSignal`, `forEachTakesADerivedSignal`), and the rest of the
suite now passes the concrete signal `makeInput` returns instead of wrapping it,
so the concrete path is exercised throughout.

`TKeyFunc` and `TDelegate` are constrained with `std::is_invocable_v`, and
`ArraySignal::map`'s function the same way. A wrong callable now reads:

```
error: no matching function for call to 'forEach'
note: candidate template ignored: requirement
      'detail::isForEachCallable<(lambda...), (lambda...), Item>' was not satisfied
```

Not `is_invocable_r`, and worth saying why: the results are exactly what the key
type and the built value type are deduced *from*, so there is nothing to give it
to compare against. Invocability is the whole of what can be checked here.

**Convertible element values — recommend against.** `T` is currently deduced
from the source and is also what the delegate is handed as `AnySignal<T>` and
what `keyFn` takes. Allowing the source's element type to differ means `T` can no
longer be deduced from the source, so every call site would have to name it
explicitly — including the many that want no conversion at all. It also needs an
extra map node converting the vector before keying, an O(n) convert per update.
The same effect is already available, explicitly and for free, by mapping the
source vector before `forEach`, which says at the call site what the conversion
is. `ArraySignal`'s own constructors do still take convertible values, since
there is no deduction to lose there.

**2. `ArrayId` kept, with the reason in the doxygen.** It is drawn from the same
counter as the `btl::UniqueId`s that key a `DataContext` and means something
else — an entry in an array node's own table. One type for both would let
`findData(element.id)`, or the reverse, compile and quietly find nothing. That is
the reason you suggested, and it is the real one; it is now one line above the
class rather than implied.

**3. `ArrayConstant` deleted.** One use site, and `signal::constant` covers it,
so it is a helper function now. The id is minted where the element comes into
existence — at construction — rather than by a node whose only job was to
allocate it at `initialize`.

That does mean the description carries the constant items' ids, which contradicts
the design's "ids are minted per context" rule, so the rule is amended rather
than the code bent: an id is minted **where its element is**. A `forEach` id has
to be per context because the key is not known until then; a constant's does not,
and an id names nothing but an entry in the per-context tables that match it. Two
contexts over one constant element look their own values up under the same id and
never meet.

MSVC and clang-cl, Debug, whole project: clean, 20 array tests, all four suites
pass. CI re-running on the new tip.

PR #115, opened 2026-07-21
Reimplement layout() over the array operators

samienne:
Step 4 of `docs/design/arraysignal.md`: `bqui`'s static layout engine is
reimplemented over `map`/`scatter`/`join`, keeping the `SizeHintMap` and
`ObbMap` signatures so `box`, `stack` and `uniformGrid` are untouched at
their call sites. Stacked on #114 (`arraysignal-scatter`).

The step was a verification, and the design held: the fan-in, the compute
and the fan-out map onto the operators with nothing added. `layout()`'s body
is now the builders as a constant `ArraySignal<AnyBuilder>`, `map` into
`join` for the hints, the caller's `ObbMap` as an ordinary map over the
joined vector, and `scatter` handing each child its obb. The dead
heterogeneous tuple path - `accumulateSizeHintsTuple` and `MapObbs` - goes
with it; a repo-wide search found no caller of either.

All 24 layout tests stay green, including the two that pin current-but-wrong
behaviour belonging to step 5 (`uniformGridSizeHintIgnoresCellSpans` and
`dynamicBoxDoesNotApplyGravity`). Verified with clang-cl and MSVC.

### Two things the design got slightly wrong, corrected in the record

- **"Per frame, both paths do the same work" was too strong.** The build
  happens once per identity, but the keyed zip feeding `scatter`'s slices
  runs on every emission, so a *resize* costs O(n log n) with n allocations
  where the old fan-out cost O(n). An idle frame is still free.
- **The step is not quite behaviour-neutral.** A child's build function now
  runs from `scatter`'s delegate, which is once per `DataContext` rather than
  once per description. Two contexts over one layout build two sets of
  children. That follows from where the array keeps its state and is the more
  correct of the two, but it is a change.

### Alignment and shared state

The two branches - the hint fan-in and, after the obbs, the element fan-in -
consume one array. Both read `array.elements()`, which is behind a `share()`,
so within a context they see one element sequence in one pass and the
aggregate is aligned with the membership by construction, exactly as
*Alignment is a promise* predicts. The per-branch instantiated-state hazard
is not reachable here: no element signal is joined twice, and with no
`forEach` in this step there is no delegate state to diverge.

`getSizes`' division by an empty interval is guarded rather than left to the
arithmetic. The infinities and the NaN it produced were clamped back to the
same answers, so the result is unchanged for every input a hint can produce.

### DIFF
--- a/src/bqui/src/widget/layout.cpp
+++ b/src/bqui/src/widget/layout.cpp
@@ -37,8 +37,9 @@ bq::signal::AnySignal<widget::Instance> buildChild(
         widget::AnyBuilder const& builder,
         bq::signal::AnySignal<avg::Obb> obb)
 {
-    auto transform = obb.map(&avg::Obb::getTransform);
-    auto size = obb.map(&avg::Obb::getSize);
+    auto shared = obb.share();
+    auto transform = shared.map(&avg::Obb::getTransform);
+    auto size = shared.map(&avg::Obb::getSize);
 
     auto placed = builder.clone()
         | modifier::transformBuilder(std::move(transform));
--- a/src/bqui/src/widget/layout.cpp
+++ b/src/bqui/src/widget/layout.cpp
@@ -10,7 +10,7 @@
 
 #include "bqui/provider/providebuildparams.h"
 
-#include <bq/signal/combine.h>
+#include <bq/signal/arraysignal.h>
 
 namespace bqui::widget
 {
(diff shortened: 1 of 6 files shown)


src/bqui/src/widget/layout.cpp
```diff
@@ -9,63 +9,74 @@
...
-    auto widget = makeWidgetWithSize(
-            [](auto size, auto obbMap, auto hintsSignal, auto builders)
-            {
-                auto obbs = merge(std::move(size), hintsSignal).map(obbMap).share();
+bq::signal::AnySignal<widget::Instance> buildChild(
+        widget::AnyBuilder const& builder,
+        bq::signal::AnySignal<avg::Obb> obb)
+{
+    auto transform = obb.map(&avg::Obb::getTransform);
+    auto size = obb.map(&avg::Obb::getSize);
```
samienne: We are duplicating the obb signal so we must make sure it is shared.

src/bqui/src/widget/layout.cpp
```diff
@@ -9,63 +9,74 @@
...
+                {
+                    return builder.getSizeHint();
+                })).share();
 
-                        ++index;
+    auto widget = makeWidgetWithSize(
+            [](auto size, auto obbMap, auto hints, auto array)
+            {
+                auto obbs = merge(std::move(size), std::move(hints))
+                    .map(std::move(obbMap));
```
samienne, line 65: OK, so here, the obbs are not shared, so we are actually calculating the obbmap twice for each widget.

samienne:
Fixed in 30c2979, and it is worth being precise about which duplication it was.

The caller's `ObbMap` was **not** running twice per widget. The aggregate is
already shared inside `scatter`, which builds
`merge(array.elements(), aggregate).map(keyByIdentity).share()`, so `ObbMap`
evaluates once per pass however many children read a slice. Adding a
`.share()` on `obbs` in `layout()` would have been a second shared node over
a single consumer, so I have not added one.

What was actually duplicated is the **per-element slice chain in
`buildChild`**: `obb` arrived unshared and was mapped twice, once for the
transform and once for the size, which instantiated `pick(slices, id)` and
its `requirePresent` check once per map. Each child therefore did two map
lookups and two `optional<Obb>` unwraps per pass instead of one. Sharing
`obb` before the split collapses that back to one, and it is the same shape
the old code had, where the two maps hung off a shared `obbs`.

I swept the rest of the rewritten `layout()` for the same pattern. Every
other signal consumed more than once already has a `share()` between it and
its consumers: `hints` is explicitly shared and feeds both `setSizeHint` and
the widget branch; `array` is consumed by the hint fan-in and by `scatter`,
but an `ArraySignal` holds its elements behind a `share()` in its own
constructor; `obbs`, `transform` and `size` each have exactly one consumer.

One pre-existing duplication I did **not** touch, since it predates this
branch and is unchanged by it: a child's own size hint signal is instantiated
both by the hint fan-in and again inside `handleGravity` when the element is
built. The old `layout()` did the same thing, and fixing it means changing
the builder plumbing rather than this function.

Layout tests all green under clang-cl.

samienne:
Both recorded in bd30bc5.

**The rule → `docs/conventions.md`**, as its own short section next to the
other signal-graph gotcha (`merge()` has no zero-argument form). It states that
a signal consumed from two places without an intervening `.share()`
instantiates and evaluates the whole chain behind it once per consumer per
pass, and that this is a defect rather than a style preference — the values
still agree, so nothing fails; the work is just done twice and anything
stateful in the chain exists twice. It then says why only review catches it:
signals were move-only with an explicit `.clone()` to opt into copying, exactly
so a second consumer could not appear by accident, and copying was enabled
later because move-only was too painful to write against. The leftover
`.clone()` methods and the habit of passing signals into widget-building
functions as parameters are named as residue of that.

**The size hint duplication → `src/bqui/AGENTS.md`**, under Traps, marked known
and unfixed. I did not use the design doc's *Future directions*: that file
says it goes away when the array work lands, and this outlives it — it is a
`layout()`/builder issue, not an `ArraySignal` one. The note says what it is
(the hint fan-in reads `getSizeHint()`, `handleGravity` reads it again when the
element is built, so the chain runs once per child per pass), and what fixing
it costs (the builder handing out one shared hint signal rather than a fresh
one per call — builder plumbing, not layout).

Stamps on both files updated to 30c2979. clang-cl rebuild clean, all 24 layout
tests green. Still draft.

PR #116, opened 2026-07-21
Delete dynamicBox and route hbox/vbox through the unified layout()

samienne:
Step 5 of the `ArraySignal` design (`docs/design/arraysignal.md`): `hbox` and
`vbox` take an `ArraySignal<AnyWidget>`, so a fixed child list and one whose
membership changes are the same case and run through one engine. `dynamicBox`
is deleted.

Both of its recorded defects go with the implementation rather than being fixed:

- **`handleGravity()`** - `layout()` puts `modifier::handleGravity()` in front of
  every child, so a gravity set on a child of a dynamic `hbox`/`vbox` is now
  honoured. It composed into `scatter`'s delegate with nothing threaded around,
  as the design predicted: the delegate has the child's own hint and gravity
  through the child's value, and the assigned outer size through the
  identity-matched obb.
- **The element cache** - there is no successor for it to be wrong in.
  `scatter` matches by identity, so nothing scans a previous result and nothing
  can discard a genuinely changed value for an existing id. The positional
  re-keying block and its `assert(obbs.size() == builders.size())` go with it;
  `scatter` checks the aggregate's arity in one place instead.

## One difference the design's audit missed

`dynamicBox` gave every child an `avg::UniqueId` through
`modifier::setElementId`; the static engine never did, and the audit under *One
layout engine* did not list it. `avg::ContainerNode::update` matches its
children **by id when they have one and by position otherwise**, so dropping it
would have meant that erasing a middle row of `testapp1`'s list animated the
*last* row out and diffed every survivor against its neighbour's previous
subtree. The unified engine mints the id in `scatter`'s delegate - once per
identity - so both paths have it. The static path pays one node per child for a
match that agrees with the positional one; that is the direction to err in,
since losing identity is a correctness regression and gaining it is a cost.
Found by the clean-context correctness review, not by me.

## One overload each

`hbox`/`vbox` take `ArraySignal<AnyWidget>` and nothing else. A
`std::vector<AnyWidget>` overload alongside it would make `hbox({a, b, c})`
ambiguous, since a braced list of widgets is also an `ArraySignal`. A caller
with a runtime-built list builds a `std::vector<ArraySignal<AnyWidget>>`
instead, which an `AnyWidget` converts into element-wise; the layout tests do
exactly that. `layout()` keeps its vector overload for `stack` and
`uniformGrid`, which are only ever called with a named vector, so those two and
`box` are otherwise untouched. Every braced call site in the tree - `scrollView`,
`testapp1` - compiles unchanged.

## `adder` had to move in the same change

This part was meant to be a later PR and could not be. Deleting `dynamicBox`
removes the only consumer of `AnySignal<std::vector<std::pair<size_t,
AnyWidget>>>`, which is `dataBind`'s exported shape, so `src/testapp1/adder.cpp`
had to move to `forEach` here. No adapter could have deferred it: turning a
signal *of widgets* into an `ArraySignal<AnyWidget>` means reading a widget out
of a signal while describing the graph, which is exactly what cannot be done and
exactly why `dynamicBox` had to either discard a changed widget or rebuild it.

`dataBind`, `dataSourceFromCollection`, `Collection`, `DataSource` and
`databindtest.cpp` are untouched and still work; `adder` simply no longer uses
`dataBind`. Three things fell out of the port, all recorded in the design doc:

- A `Collection` reaches `forEach` through an ordinary fold that re-reads the
  collection on every event rather than replaying it - a few lines against
  `dataBind`'s branch per event kind, and nothing is lost by the coarseness
  because `forEach` keys by the collection's own id.
- **`forEach` does not hand its delegate the key**, so a delegate that needs the
  item's identity has to re-derive it from the item signal and produce its
  callables through a `map`. `dataBind`'s delegate took the id directly, so this
  is a real ergonomic regression at the one call site that exists. It is not the
  rejected *index* - a key is stable identity - and giving the delegate its key
  is a small change to `forEach` that should be decided on its own. Not made
  here.
- A value change now reaches every item rather than one, because a `pick` is a
  plain map over the one shared keyed source. Correct, and O(rows) per event
  rather than O(1); `.check()` at the point of use is the documented answer and
  no call site needs it yet.

## Tests

The four `dynamicBox` tests are reworked onto the unified path rather than
dropped, driving membership through `forEach` over an input vector of probe
indices. The probe widget is now built entirely out of its index signal, which
is what a `forEach` delegate is actually given.

- `dynamicBoxDoesNotApplyGravity` -> `dynamicHboxAppliesGravity`, and it flips:
  the probe asking for 50x30 in a 50x100 slot is now centered at y = 35 rather
  than stretched. This is the headline behavioural change.
- `dynamicHboxBuildsEachChildOncePerIdentity` is new. It counts the times the
  *container* asks a child to build itself - not the times the delegate ran,
  which would say nothing about the layout - across an insert and a removal, and
  pins that a survivor is not rebuilt. It derives the per-child cost from the
  first pass rather than hard-coding it, because `handleGravity()` builds each
  child twice.
- `uniformGridSizeHintIgnoresCellSpans` is unmoved and still green, as expected:
  it is a latent bug in `uniformGrid`'s own `SizeHintMap` and nothing here goes
  near it.

Every other layout test is unchanged in expectation.

## Verified

`lw build` + `lw test` green on `Debug:clang-cl-18.1.7`,
`Debug:msvc-17-2022-enterprise` and `Release:clang-cl-18.1.7`; no new warnings.


src/bqui/src/widget/layout.cpp
```diff
@@ -111,5 +99,21 @@ widget::AnyWidget layout(SizeHintMap sizeHintMap,
...
+    std::vector<bq::signal::ArraySignal<widget::AnyWidget>> children;
+    children.reserve(widgets.size());
+
+    for (auto&& widget : widgets)
+        children.push_back(std::move(widget));
+
+    return layout(
+            std::move(sizeHintMap),
+            std::move(obbMap),
+            bq::signal::ArraySignal<widget::AnyWidget>(std::move(children))
```
samienne, line 118: Shouldn't this already accept std::vector<AnyWidget> without needing to move the widgets into children? This would basically celar all the unnecessary moving.

PR #118, opened 2026-07-21
Make App's window list an ArraySignal

samienne:
Reworks PR #99 onto `ArraySignal`. Based on `arraysignal-scatter` (PR #114), not
`master`; the old branch predates the design it should be built on, and its
hand-written by-id reconcile is what `forEach` replaces.

## The design question this settles

`docs/design/arraysignal.md` asked, as an open question, whether any identity
surfaces at the exit, and named this PR as the consumer that would decide it.

**It does not.** `App` takes a `bq::signal::ArraySignal<Window>` and builds one
`WindowGlue` per identity with `map()`. The array *is* the reconciliation:
create on arrive, destroy on depart, survivors untouched. The `WindowList`
alias, the producer-assigned `size_t`, and the reconcile loop in `App::run` are
all deleted, and no id reaches user code.

What the doing of it taught, now recorded in the design:

- The consumer's `map` is the reconcile. Nothing else was needed.
- A value that is not a signal leaves the array as `constant(value)`. That is
  what the exit is for — `join`'s Doxygen now says so.
- The exit hands back a plain vector, and everything `App` did with ids is a
  fold over it.

The one thing a caller gives up is having a single `forEach` build differently
shaped items, since the delegate gets the item's value and not its key. The
answer is one array per shape, concatenated — `testapp1` does exactly that for
its second window, so **the keyed delegate on `foreach-keyed-delegate` was not
needed and is not cherry-picked.**

`onWindowsReconciled` is **gone, with nothing in its place.** The list is the
source of truth for which windows exist, so a change notification on `App` is
the wrong shape: what a window offers is wired back into whatever the list is
built from, and the removal flows through the array. A window's `onClose`
removes its own key; so does a button in its own UI. `testapp1` shows both
halves on its second window.

That also removes the reason to argue about the payload — reporting titles was
doubly wrong, since a title is presentation, not identity. The finding
generalises: a consumer that thinks it needs a change notification from the
array is usually holding a second copy of the membership.

## A constraint the design did not state, found by breaking it

**An element's signal may not be read on a clock of its own.** *The pick
construction* requires a pick to be dropped no later than the update that
observes its key leaving. A consumer that only reads elements inside the array's
own update pass gets that for free — the unified layout engine will be one,
which is why the obligation looked like a formality.

`App` was not one. Each `WindowGlue` holds its window's title and widget signals
in its own `SignalContext`, updated from OS input handlers, and an `Input` is
global to every context. So a window with its own close button removed its key
and then updated a signal naming that key, `requirePresent` threw out of the
event loop, and the process died. Reproduced in `testapp1` by clicking, not
reasoned about.

The fix is not to soften `requirePresent` — a latch there hands back a stale
value, which the design already rejects. It is to give the element's signals one
clock: input handlers now request a frame instead of transacting inline, so
per-window updates happen after the array has been reconciled and a departed
window's glue is gone before anything can touch it. The rule is now in the
design doc and in `src/bqui/AGENTS.md`.

A second ordering rule came out of review: a glue is released only after the
main render queue has caught up, because an in-flight frame holds the window's
framebuffer, which holds the window by reference.

## The multi-window defect: half real, and fixed

Rendering and input in an already-open window are **fine** when a second window
opens at runtime — verified by running `testapp1`. The first window keeps
animating and keeps accepting clicks with a second window open.

What was broken is the other direction. `WglWindow` had **no destructor**, so
releasing an `ase::Window` left its `HWND` alive: a closed window stayed on
screen, painted nothing, and answered no input. Two open/close cycles left three
"Second window" windows on screen. `GlxWindow` has always destroyed its X
window; this is the WGL half, and nothing could close a window mid-run before
this PR, which is why it had gone unnoticed.

## The old CI failures were not staleness

PR #99's Linux and Windows legs failed with `Unable to open X display` and
`Unable to get wglCreateContextAttribsARB.` — `App::run` opens real windows, and
CI has no display and no GPU context. macOS passed because it builds the dummy
backend. That is structural, not stale.

So `ase` now exports `ase_is_headless` and `bqui` compiles `test/apptest.cpp`
only where the headless backend is built. The app loop is otherwise verified by
running the app.

## Verified

- All twelve CI legs green, plus `ci-success`.
- `clang-cl` Debug + Release and `msvc-17` Release locally: build and tests
  green.
- The headless backend forced on locally so `apptest.cpp` compiles and runs;
  both tests pass. They also run for real on the macOS legs. Their assertions
  were rewritten after review to be able to fail: a window's title is read once
  per glue built, so counting reads observes an opening and catches a rebuild,
  and both tests cap their frames so a stalled script fails instead of hanging.
- `testapp1` driven with synthetic input through: open the second window, close
  it from its **own title bar**, reopen, close it from its **own button**, then
  close the main window and confirm that ends the app. Screenshots of the main
  window at each step show it still animating and still interactive.

## Known gaps, and why the animation one does not block this

Making self-removal the idiom raises the obvious question of whether the
`AnimationGuard` gap is now on the main path. **It is not, and here is the
reason rather than the assertion:** the close path never transacts at all.
`WindowGlue`'s close callback only invokes the window's own callbacks and
returns — no `makeTransaction`, no signal evaluation — so removing a window from
its `onClose` is safe *by construction*, not by the frame-phase rule. The other
half of the idiom, a button inside the window, is safe by that rule. Neither
opens an animation scope.

Reaching the `AnimationGuard` gap therefore takes an author explicitly wrapping
a *removal* in `withAnimation`, and there is no window-level exit animation in
the toolkit that would make that routine — `withAnimation` marks widget signal
changes, and a window being destroyed has no render tree left to animate.
Verified by running: title-bar close, widget-button close, reopen, and quit, all
clean.

What the fix would need, so it can be scoped as its own PR: `AnimationGuard`
must stop being a pair of synchronous transactions of every glue. Two shapes
work — carry the options into the frame phase and let the next per-window
transaction apply them (loses the commit-before boundary for changes made
earlier in the same event drain), or give `App` a reconcile-and-defer-destruction
step the guard runs first, so departed glues leave the live set without being
destroyed under their own handler (keeps the semantics). Both need state that
today lives as locals in `App::run` moved into `AppDeferred`, and both change
observable animation behaviour, so they want their own run-the-app verification.

The second gap is unchanged: a list changed from *within* the frame phase — one
window's update removing another's key — is reconciled only on the next pass.

Both are recorded in `src/bqui/AGENTS.md` and the design doc.

Destroying a WGL window mid-run also leaves two loose ends in the render
context that were unreachable while windows lived for the whole process: the
dispatched context caches the last DC it made current and nothing invalidates
it, and with no window left the loop keeps submitting fences against a context
current on a destroyed DC. Noted in `src/ase/AGENTS.md`.

## Known open points for the reviewer

- Gating `apptest.cpp` on `ase_is_headless` means it compiles only on macOS. A
  build option that forces the dummy backend for the test binary would decouple
  the two.
- `App::run()`'s close wiring now reaches windows opened later, so closing any
  window ends the app. That is the pre-existing meaning of the no-argument
  `run()`, applied to a list that can grow, and it is now said in its Doxygen —
  a list whose windows come and go wants `run(running)`. `testapp1` switched to
  that so its two windows can close to different effect.

Both clean-context reviews were run again after the observer removal and their
findings applied.

Draft: the whole stack is gated pending review.

### DIFF
--- a/src/bqui/include/bqui/app.h
+++ b/src/bqui/include/bqui/app.h
@@ -3,12 +3,15 @@
 #include "window.h"
 #include "bquivisibility.h"
 
+#include <bq/signal/arraysignal.h>
 #include <bq/signal/signal.h>
 
 #include <avg/curve/curves.h>
 
 #include <btl/shared.h>
 #include <btl/visibility.h>
+
+#include <optional>
 
 namespace bqui
 {
@@ -20,9 +23,31 @@ namespace bqui
     public:
         explicit App();
 
-        App windows(std::initializer_list<Window> windows) &&;
+        /** @brief Sets the windows the app opens.
+         *
+         * The list is the only thing that says which windows exist. A braced
+         * list is a constant array, so `windows({ a, b })` opens two windows
+         * and never changes; a list built with bq::signal::forEach() opens and
+         * closes windows as its keys come and go, and every window that stays
+         * keeps everything it had — its widgets, their state, and its own OS
+         * window.
+         *
+         * A window closes by leaving the list, so wire what a window offers to
+         * whatever the list is built from: `Window::onClose` on a window whose
+         * key comes from an input removes that key.
+         */
+        App windows(bq::signal::ArraySignal<Window> windows) &&;
 
+        /** @brief Runs until `running` is false. */
         int run(bq::signal::AnySignal<bool> running) &&;
+
+        /** @overload
+         *
+         * Runs until a window closes — **any** window, including one that
+         * opened later. That is what a single-window app wants; a list whose
+         * windows come and go wants the overload above, so that closing one
+         * window can mean removing it rather than stopping.
+         */
         int run() &&;
 
         [[nodiscard]]
--- a/docs/design/arraysignal.md
+++ b/docs/design/arraysignal.md
@@ -16,7 +16,7 @@
 > `docs/conventions.md`, the settled *why* to `docs/decisions.md`, and the API
 > contract to Doxygen in the new headers — and this file goes away.
 
-*Last verified against `916c25b` (2026-07-21).*
+*Last verified against `90c50f2` (2026-07-21).*
 
 ## The problem
 
--- a/docs/design/arraysignal.md
+++ b/docs/design/arraysignal.md
@@ -1,9 +1,10 @@
 # Design: `ArraySignal<T>`, `forEach`, and one layout engine
 
-> **Status: revised design, implementation started.** Steps 0 to 3 of
-> *Implementation order* have landed; everything in `bqui` is
-> outstanding, and this document has been amended to describe what was built
-> rather than what was first proposed. The earlier implementation on PR #100 is
+> **Status: revised design, implementation started.** Steps 0 to 4 of
+> *Implementation order* have landed — all of `bq`, plus `App`'s window list;
+> the layout rework is outstanding. This document has been amended to describe
+> what was built rather than what was first proposed. The earlier
+> implementation on PR #100 is
 > a **superseded spike**:
 > it put the per-identity table in the description rather than in the
 > `SignalContext`, which is a correctness defect, not a limitation. This revision
@@ -15,7 +16,7 @@
 > `docs/conventions.md`, the settled *why* to `docs/decisions.md`, and the API
 > contract to Doxygen in the new headers — and this file goes away.
 
-*Last verified against `86ddac5` (2026-07-21).*
+*Last verified against `916c25b` (2026-07-21).*
 
 ## The problem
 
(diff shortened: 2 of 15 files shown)


src/bqui/include/bqui/app.h
```diff
@@ -20,7 +25,29 @@ namespace bqui
...
+         * which the window list changed — one opened, one closed, or the order
+         * moved — from the run loop's own thread.
+         * Titles are what names a window outside the process, which is what an
+         * introspecting test or tool has to go on; a window carries no
+         * identity of its own, and the identity that keeps it alive across an
+         * edit is internal to its array.
+         */
+        App onWindowsChanged(
+                std::function<void(std::vector<std::string> const&)> callback
+                ) &&;
```
samienne: I don't like this callback. The idea is to manage a collection of windows. The windows themselves have callbacks (now only for close) and that should be wired to remove the window from the list.

samienne:
Closing in favour of #121, which now carries this change in full — these five commits are the first five there, unchanged, not a squash or a re-implementation.

The two were separate because they were written in sequence, but they are one change to read: this PR makes `App`'s window list an `ArraySignal<Window>`, and #121's second half puts an App-owned collection on top of it. Reviewing them apart means reviewing an API in its intermediate form — the one where every caller must build and manage an array — and then reviewing its replacement. Together, the question is the one that actually matters: whether the general mechanism and the friendlier surface over it are the right pair, and whether they coexist without two sources of truth.

The stack was reordered at the same time. `sharedvector` (#113) is `bq` and the window work is `bqui`, so it now sits below rather than above: `arraysignal-scatter` → `sharedvector` → #121.

Everything recorded here survives in #121's description: the identity question this PR settled, the constraint it discovered about reading an element's signal on a clock of its own, the WGL `HWND` leak it uncovered, the headless-only CI split for `apptest.cpp`, and the two known gaps in `AnimationGuard` and in list edits made from within the frame phase — which #121 narrows to the caller-supplied path only.

PR #121, opened 2026-07-21
Let App own its windows via decoupled value handles

samienne:
App owns its windows imperatively. `Window` is a copyable value handle; the widget is separate, supplied at mount.

## Window
- `window(title)` mints a handle with a fresh identity, belonging to no app, holding no widget.
- Copyable/movable; copies share one identity.
- API: `onClose(cb)`, `getTitle()`, `getId()`, `close()`, `clone()`.
- Backed by a shared `WindowData` (id, title, close callbacks, weak app ref); survives remove then re-add.

## App
- `addWindow(window, widget)` opens and mounts; the widget is content, not identity.
- `removeWindow(id)` / `Window::close()` unmount: destroy the widget, keep the data.
- Re-adding a removed window supplies a fresh widget.
- `getWindows()` snapshot; `getWindowsSignal()` reactive.
- `run(running)` runs until `running` is false; `run()` runs until the last window closes.
- `addWindow`/`removeWindow`/`getWindows`/`Window::close` are lock-guarded (usable from a worker thread); everything else is the app thread.
- Throws `std::invalid_argument` when adding a window already open here or in another app (a window and its copies are one window).

## Internals
- `WindowImpl` (App-owned, keyed by id) owns the OS window, painter, and widget; in `src/bqui/src/windowimpl.h`. `app.cpp` is the collection + run loop.
- No retain cycle: App -> WindowImpl -> widget -> Window -> WindowData, and WindowData -> app is weak. A close button capturing its own Window is safe; `apptest` asserts teardown.

## Behaviour change
- `run()` (no arg) stops at the last window close, not the first close (see `docs/decisions.md`). `run(running)` unchanged.

## Tests
- `windowtest.cpp`: backend-free; covers the handle and `close()` after the app is destroyed, on every platform.
- `apptest.cpp`: run loop; builds only where `ase` is headless (macOS CI).

`bq` untouched. Commits: code, then docs.

### DIFF
--- a/src/bqui/include/bqui/app.h
+++ b/src/bqui/include/bqui/app.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "window.h"
+#include "widget/widget.h"
 #include "bquivisibility.h"
 
 #include <bq/signal/signal.h>
@@ -8,7 +9,11 @@
 #include <avg/curve/curves.h>
 
 #include <btl/shared.h>
+#include <btl/uniqueid.h>
 #include <btl/visibility.h>
+
+#include <optional>
+#include <vector>
 
 namespace bqui
 {
@@ -20,10 +25,56 @@ namespace bqui
     public:
         explicit App();
 
-        App windows(std::initializer_list<Window> windows) &&;
+        /** @brief Opens a window the app owns, mounting the given widget in it.
+         *
+         * The widget becomes the window's mounted content, not part of its
+         * identity: removing the window destroys the widget, and adding the same
+         * window again re-supplies a fresh one. The collection is imperative, so
+         * a window may be added and removed both before and while the app runs.
+         *
+         * @throws std::invalid_argument if the window is already open, here or
+         *         in another app. A window and its copies are one window, so
+         *         adding a copy of an open window is adding it twice. A removed
+         *         window belongs to no app again and may be opened anywhere.
+         */
+        App& addWindow(Window window, widget::AnyWidget widget);
 
-        int run(bq::signal::AnySignal<bool> running) &&;
-        int run() &&;
+        /** @brief Closes the app's window with this identity.
+         *
+         * Does nothing if no window in the collection has it. The window's own
+         * data outlives this whenever a Window naming it is still held; only its
+         * mounted widget is torn down.
+         */
+        void removeWindow(btl::UniqueId id);
+
+        /** @brief The app's windows as they are right now.
+         *
+         * The snapshot form, for code outside a signal graph — counting the
+         * open windows, or finding one to remove.
+         */
+        std::vector<Window> getWindows() const;
+
+        /** @brief The app's windows, as a signal.
+         *
+         * The reactive form of the same collection, for a UI that follows it.
+         */
+        bq::signal::AnySignal<std::vector<Window>> getWindowsSignal() const;
+
+        /** @brief Runs until 'running' is false, whatever the windows do.
+         *
+         * The calling thread is the app's thread: run() and withAnimation()
+         * belong to it, and every window is built, drawn and driven there. The
+         * window collection is the exception — addWindow(), removeWindow(),
+         * getWindows() and Window::close() are lock-guarded, so a worker thread
+         * can open or close a window itself.
+         */
+        int run(bq::signal::AnySignal<bool> running);
+
+        /** @overload
+         *
+         * Runs until the last window is closed.
+         */
+        int run();
 
         [[nodiscard]]
         AnimationGuard withAnimation(avg::AnimationOptions options);
@@ -31,6 +82,8 @@ namespace bqui
         friend class AnimationGuard;
 
     private:
+        int runUntil(bq::signal::AnySignal<bool> running);
+
         inline AppDeferred* d()
         {
             return deferred_.get();
--- a/src/bqui/include/bqui/window.h
+++ b/src/bqui/include/bqui/window.h
@@ -1,20 +1,32 @@
 #pragma once
 
-#include "widget/widget.h"
-
 #include "bquivisibility.h"
 
 #include <bq/signal/signal.h>
 
-#include <btl/cloneoncopy.h>
+#include <btl/uniqueid.h>
+
+#include <functional>
+#include <memory>
+#include <string>
 
 namespace bqui
 {
+    class WindowData;
+    class App;
+    class AppDeferred;
+    class WindowImpl;
+
+    /** @brief A window in the application.
+     *
+     * Create a Window, add it to the App with the widget to show in it, and
+     * close it when you are done.
+     */
     class BQUI_EXPORT Window
     {
     public:
-        Window(widget::AnyWidget widget,
-                bq::signal::AnySignal<std::string> const& title);
+        /** @brief Mints a window with a fresh identity, belonging to no app. */
+        explicit Window(bq::signal::AnySignal<std::string> const& title);
 
         Window(Window const&) = default;
         Window& operator=(Window const&) = default;
@@ -22,13 +34,22 @@ namespace bqui
         Window(Window&&) = default;
         Window& operator=(Window&&) = default;
 
+        /** @brief Adds a callback run when the window's title bar closes it. */
         Window onClose(std::function<void()> const& cb) &&;
-
-        widget::AnyWidget getWidget() const;
 
         bq::signal::AnySignal<std::string> const& getTitle() const;
 
         void invokeOnClose() const;
+
+        /** @brief This window's identity, which its copies share. */
+        btl::UniqueId getId() const;
+
+        /** @brief Removes this window from the app that owns it.
+         *
+         * A no-op if the window is in no app, or if that app is gone. Closing
+         * twice is a race a UI can lose without anything being wrong.
+         */
+        void close() const;
 
         Window clone() const
         {
@@ -36,12 +57,18 @@ namespace bqui
         }
 
     private:
-        widget::AnyWidget widget_;
-        bq::signal::AnySignal<std::string> title_;
-        std::vector<std::function<void()>> closeCallbacks_;
+        friend class App;
+        friend class AppDeferred;
+        friend class WindowImpl;
+
+        std::shared_ptr<WindowData> const& data() const
+        {
+            return data_;
+        }
+
+        std::shared_ptr<WindowData> data_;
     };
 
-    BQUI_EXPORT auto window(bq::signal::AnySignal<std::string> const& title,
-            widget::AnyWidget widget) -> Window;
+    /** @brief Mints a window with the given title. */
+    BQUI_EXPORT Window window(bq::signal::AnySignal<std::string> const& title);
 }
-
--- a/src/bqui/meson.build
+++ b/src/bqui/meson.build
@@ -80,13 +80,27 @@ libbquidep = declare_dependency(
   include_directories: include_directories('include')
   )
 
-bquitest =  executable('bquitest',
+bquitestsrc = [
   'test/collectiontest.cpp',
   'test/databindtest.cpp',
   'test/layouttest.cpp',
   'test/shapetest.cpp',
   'test/widgetinstancemodifiertest.cpp',
   'test/widgettest.cpp',
+  'test/windowtest.cpp',
+  ]
+
+# App::run opens real windows, so it can only be exercised where ase builds its
+# headless backend. This gate is the minimal thing needed to run apptest.cpp
+# against ase's dummy backend; the proper headless platform (env-selected and
+# frame-stepped) arrives in a follow-up (PR #96), and this should be revisited
+# to use it once that lands.
+if ase_is_headless
+  bquitestsrc += 'test/apptest.cpp'
+endif
+
+bquitest =  executable('bquitest',
+  bquitestsrc,
   dependencies: [gtestdep, libbquidep],
   )
 
--- a/src/bqui/src/app.cpp
+++ b/src/bqui/src/app.cpp
@@ -1,13 +1,16 @@
 #include "bqui/app.h"
 
 #include "bqui/window.h"
-#include "bqui/send.h"
+#include "bqui/modifier/background.h"
 
 #include "debug.h"
+#include "windowdata.h"
+#include "windowimpl.h"
 
 #include <bq/signal/input.h>
 #include <bq/signal/updateresult.h>
 #include <bq/signal/signalcontext.h>
+#include <bq/signal/sharedvector.h>
 
 #include <avg/rendertree.h>
 #include <avg/painter.h>
@@ -33,28 +36,24 @@
 
 #include <tracy/Tracy.hpp>
 
+#include <algorithm>
 #include <chrono>
+#include <functional>
+#include <mutex>
+#include <stdexcept>
 #include <unordered_map>
 #include <memory>
 #include <optional>
+#include <string>
+#include <utility>
+#include <vector>
 #include <cstdint>
 
 namespace bqui
 {
 
-namespace
-{
-    uint64_t s_frameId_ = 0;
-
-    uint64_t getNextFrameId()
-    {
-        return ++s_frameId_;
-    }
-} // anonymous namespace
-
-class WindowGlue;
-
-class BQUI_EXPORT AppDeferred
+class BQUI_EXPORT AppDeferred :
+    public std::enable_shared_from_this<AppDeferred>
 {
 public:
     AppDeferred()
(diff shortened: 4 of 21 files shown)


samienne:
Both clean-context reviews ran and their findings are addressed in the two follow-up commits.

**Style review** — the `apptest.cpp` preamble claimed `App` reports nothing about its windows, which this change made false; a comment carried the behaviour change that `docs/decisions.md` already records; `src/bq/AGENTS.md`'s stamp was not bumped; the collection tests now have their own gtest suite. The one substantive item was `runStopsWhenTheLastWindowCloses` waiting out 50 ms to decide the loop had not stopped — a wall-clock ordering proxy, which `docs/decisions.md` explicitly moved away from. It now waits for the loop to *open a window added after the first close*, which is the same claim made against an event the app produces. The deadline that remains is a backstop against a hang, not a proxy.

**Correctness review** — three real defects:

- **The glues outlived the render context they are made of** when a run ended by exception. `windowGlues_` belongs to the app, not to the loop, so only the normal path cleared it; a throw out of `makeSignalContext` or `glues.update` — reachable from a caller-supplied array with a duplicate key — left `ase::Window`s and framebuffers alive against a destroyed platform, for `AnimationGuard` or a second `run()` to walk. A scope guard now releases them however the run ends.
- **`addWindowArray` after `run()` started silently opened nothing.** The arrays are joined once; the append went to a vector nothing would read again. Rejected with `std::logic_error` now, which also makes the asymmetry with `addWindows` ("windows may be added while the app runs") explicit rather than a trap.
- **A second OS close event re-ran a window's `onClose` callbacks.** The glue outlives the removal by up to a frame. Under the old design the callback *was* the removal mechanism, so a repeat was self-limiting; now it is not, and the glue latches.

Two sharp edges also closed: a window could be added to a second app while open in the first, leaving it open in one and closable only from the other — `add` rejects a window whose handle names a live list, and `remove` clears the handle so a closed window can be opened anywhere again. And `close()` on a window that is not there opened a write scope, which publishes a deep copy of every window that is; it checks under a read scope first.

The review also confirmed the two claims this PR rests on, independently: the two-argument `forEach` builds no pick, so an app-owned window's glue reads nothing a departing key can invalidate and `AnimationGuard` on it is safe; and the lock order between `WindowList` and `WindowHandle` has no cycle, because `close()` releases the handle mutex before it touches the list.

One thing I corrected rather than fixed: the comment on `WindowList::remove` claimed no caller-written code runs under the lock. Hoisting the departing windows out achieves that for *their* destruction, but publication copies every surviving `Window` — and a `Window` holds an `AnyWidget`, so that is a deep copy of the widget closure — and releases the previously published copy, both under the lock. That is `SharedVector`'s documented contract rather than a defect here, so the comment now says what it actually delivers instead of overstating it.

Re-verified after the fixes: clang-cl build and unit suite green, and `testapp1` driven again end to end — caller-driven window closed from its own button with the app continuing, app-owned window opened, closed by its own animating button, reopened, closed via its title bar with the app continuing, and the last window closing exited the process 0.

samienne:
Addressed. You were right on both counts: the label was false, and the operator was avoidable.

**Which case I am in: it can.** `WindowGlue` uses the window value in exactly three places — `getWidget()` (line 135), `getTitle()` (139), and the close callbacks (162-163) — and `platform.makeWindow()` takes a size, not any window data. All three are satisfiable from a value read once at construction. So the glue now takes `AnySignal<Window>`, holds it in a `SignalContext`, evaluates it once in the constructor to produce `window_`, and never updates it. There is nothing later to read — an element cannot vary at a fixed identity — so the departed-key read never happens. That is the guarantee the freeze was reaching for, obtained without an operator.

I verified the constraint #118 fixed covers this path: `WindowGlue`'s input handlers already `requestFrame()` rather than transact inline, and the one read of the window signal is at construction, before any handler can run. The App-owned path has no per-frame read of an array element signal at all.

`WindowList` now keys its `SharedVector` with `forEach(sig, &getId-lambda, [](AnySignal<Window> w){ return w; })` — the identity delegate, elements are `AnySignal<Window>` as designed. A caller-supplied array is lifted to that same shape with `.map(...)` to `constant`, so `App::run` joins one array.

The two-argument operator, its tests, and the `bq/AGENTS.md` note are gone; `git diff origin/sharedvector -- src/bq/` is empty. Docs in `docs/design/arraysignal.md` and `src/bqui/AGENTS.md` that described the frozen form now describe reading-once-vs-driving instead, and state plainly that the freeze belongs to a consumer that knows its elements don't vary, not to the vocabulary.

Re-verified by running `testapp1`: opened an app-owned window, closed it with its own animating button, reopened, closed it via its title bar with the app continuing, and closed the last window — the process exited. The window shows its title and content throughout, which is the read-once path working.

samienne:
Clean-context style review of the operator-removal commit surfaced one thing worth flagging that belongs to **#113, not this PR** — reporting rather than fixing, since #113 is approved and merging and `bq` must stay untouched here.

`src/bq/AGENTS.md`'s stamp reads `*Last verified against d2e8954 (2026-07-13)*`, but that file did not exist at `d2e8954` — it was first added on 2026-07-21, eight days later (`git cat-file -e d2e8954:src/bq/AGENTS.md` fails). So the stamp names a commit that predates the file's own creation.

This is `origin/sharedvector`'s content verbatim: `git diff origin/sharedvector -- src/bq/` is empty on this branch. The review only saw it because my operator-removal commit reverted `bq` to `origin/sharedvector`, and my earlier (now-removed) window-side edit had temporarily bumped that stamp — so the revert shows up in the diff as `e2f0ee7 → d2e8954`. Net state matches #113 exactly; I did not fix it because doing so would modify #113's approved content and break the "bq untouched by this PR" property.

Two options, your call: fix it in #113 before it merges, or leave it and I'll correct it in a later bqui-side pass once #113 has landed.

Everything else in the review came back clean or was addressed: no dangling references to the removed operator remain (the two surviving "no delegate" mentions are about the rejected bare-`AnySignal<T>` constructor, a different concept); the `WindowGlue::windowSignal_` comment is a timeless invariant with no history or doc-pointer; formatting is within 80 columns for all added code; the one over-long Markdown heading I tightened. The stamps on the three docs this PR *does* own (`arraysignal.md`, `bqui/AGENTS.md`, `decisions.md`, plus `ase/AGENTS.md`) were pointing at the pre-restack hashes and now point at the current base `77e391f`.

src/bqui/src/app.cpp
```diff
@@ -457,49 +548,160 @@ class WindowGlue
...
+        mainQueue.finish();
+
+        // Swap and then clear, rather than assign: destroying a window runs
+        // event handlers that reach back into the live set, which must be the
+        // new one by then.
+        d()->windowGlues_.swap(departing);
+        departing.clear();
+    };
+
+    collect();
```
samienne: Is the lambda just for organization or is there some other reason we create a lamdba we call immediately?

src/bqui/src/app.cpp
```diff
@@ -457,49 +548,160 @@ class WindowGlue
...
+    using RunningContext =
+        bq::signal::SignalContext<bq::signal::AnySignal<bool>>;
 
-            return running.evaluate<0>().get<0>();
-        });
+    std::optional<RunningContext> running;
+    if (runningSignal)
+        running.emplace(*runningSignal);
+
+    if (running || !d()->windowGlues_.empty())
```
samienne: The application should be runnable without any windows. I would want the default behavior to be that the application closes when there are no windows left but it should be users choice. I would tie this by mapping the window list signal into the running signal outside.

src/bqui/src/app.cpp
```diff
@@ -457,49 +548,160 @@ class WindowGlue
...
     auto startTime = clock.now();
 
     DBG("Reactive running...");
 
-    auto running = bq::signal::makeSignalContext(runningSignal);
-    platform.run(context, [&](ase::Frame const& aseFrame) -> bool
-        {
-            bq::signal::FrameInfo frame{ getNextFrameId(), aseFrame.dt };
-            running.update(frame);
+    using RunningContext =
```
samienne: This could be tied with the glues signal context. We would then have app specific signal context and a signal context per window.

src/bqui/src/app.cpp
```diff
@@ -457,49 +548,160 @@ class WindowGlue
...
+
+        // A frame that drew a departing window can still be in flight, and it
+        // holds that window's framebuffer, so nothing may release a glue until
+        // the queue has caught up.
+        mainQueue.finish();
+
+        // Swap and then clear, rather than assign: destroying a window runs
+        // event handlers that reach back into the live set, which must be the
+        // new one by then.
+        d()->windowGlues_.swap(departing);
```
samienne: It seems that all the operations on windowGlues_ are purely inside the run function so I was wondering if it should be a local variable instead of class variable.

src/testapp1/main.cpp
```diff
@@ -69,6 +71,32 @@ std::vector<std::pair<std::string, avg::Curve>> curves = {
...
+            widget::button("Close me",
+                    bq::signal::constant(std::function<void()>(
+                            [handle]()
+                            {
+                                // Animating and removing in one handler is the
+                                // combination that transacts a departing
+                                // window synchronously, so this is where it is
+                                // exercised.
+                                auto a = withAnimation(0.3f,
+                                        avg::curve::easeOutCubic);
```
samienne: I don't think animations have anything to do with closing the windows. The animations only affect the rendertree and the animations live there. With windows being closed, the window is closed immediately.

src/bqui/meson.build
```diff
@@ -80,13 +81,24 @@ libbquidep = declare_dependency(
...
   'test/widgetinstancemodifiertest.cpp',
   'test/widgettest.cpp',
+  'test/windowtest.cpp',
+  ]
+
+# App::run opens real windows, so it can only be exercised where ase builds its
+# headless backend.
+if ase_is_headless
+  bquitestsrc += 'test/apptest.cpp'
+endif
```
samienne, line 100: We have to make note that the headless pr is incomingi after this so let's not do too much for the headless setup.

src/bqui/src/app.cpp
```diff
@@ -52,38 +61,69 @@ namespace
...
+    // The app's window collection: one imperative, thread-safe, observable set
+    // of open windows, read by the run loop each frame. getWindowsSignal
+    // exports it for a UI that follows the set; nothing derives the windows
+    // from that signal.
+    bq::signal::SharedVector<Window> windows_;
+
+    // Windows added but not yet mounted, each with its widget. An OS window is
+    // built on the app thread, so addWindow (callable from any thread) only
+    // enqueues here; the run loop's sync drains it and builds the impls.
+    // Guarded because addWindow and the drain run on different threads.
```
samienne: Let's not include these details here. It will be a maintenance burden and has a large probability to get stale.

src/bqui/src/app.cpp
```diff
@@ -52,38 +61,69 @@ namespace
...
+
+    std::vector<Window> getWindows() const
+    {
+        return *windows_.read();
+    }
+
+    // The app's window collection: one imperative, thread-safe, observable set
+    // of open windows, read by the run loop each frame. getWindowsSignal
+    // exports it for a UI that follows the set; nothing derives the windows
+    // from that signal.
```
samienne: Let's remove these comments.

src/bqui/src/app.cpp
```diff
@@ -52,38 +61,69 @@ namespace
...
+    // built on the app thread, so addWindow (callable from any thread) only
+    // enqueues here; the run loop's sync drains it and builds the impls.
+    // Guarded because addWindow and the drain run on different threads.
+    std::mutex pendingMutex_;
+    std::vector<std::pair<Window, widget::AnyWidget>> pendingMounts_;
+
+    // The app's live impls, one per mounted window. The run loop syncs this
+    // each frame: mounting a pending widget for any window not yet here,
+    // tearing down any impl whose window has left. AnimationGuard walks it to
+    // transact every open window. App-thread-only.
```
samienne: Again, unnecessary implementation details that can get stale later.

src/bqui/src/app.cpp
```diff
@@ -432,7 +492,12 @@ class WindowGlue
     pmr::memory_resource* memory_;
     ase::Window aseWindow;
     ase::RenderContext& context_;
-    Window window_;
+
+    // The persistent data of the window this impl drives. The impl owns the
+    // window's contents (its widget, below) but not its identity: the data
+    // outlives the impl whenever a Window naming it is still held. Everything
+    // the impl drives per frame is the window's own title and widget signals.
```
samienne: Remove these comments.

src/bqui/src/app.cpp
```diff
@@ -457,75 +522,285 @@ class WindowGlue
...
+
+    if (app)
+        app->removeWindow(id_);
+}
+
+void AppDeferred::removeWindow(btl::UniqueId id)
+{
+    // Nothing to publish if the window is not here, and publishing is a copy
+    // of every window that is. Losing the race against a concurrent removal
+    // only costs one such copy, so the check needs no more than a read scope.
```
samienne: Remove comments

src/bqui/src/app.cpp
```diff
@@ -457,75 +522,285 @@ class WindowGlue
...
+                    return window.getId() == id;
+                });
+
+        if (found == handle->end())
+            return;
+    }
+
+    // Declared before the write scope so that it outlives it: a departed
+    // window's data may run user code as it is released, which must not run
+    // under a lock that the same code might try to take.
```
samienne: Unnecessary details that are obvious from the code.

src/bqui/src/app.cpp
```diff
@@ -457,75 +522,285 @@ class WindowGlue
...
+        departing.insert(departing.end(),
+                std::make_move_iterator(departed),
+                std::make_move_iterator(handle->end()));
+
+        handle->erase(departed, handle->end());
+    }
+
+    // A window that has left belongs to no app again, so it can be opened in
+    // one again. Its impl is torn down by the run loop's next sync, which sees
+    // the window gone from the collection.
```
samienne: Remove

src/bqui/src/app.cpp
```diff
@@ -457,75 +522,285 @@ class WindowGlue
...
+                    "know which app to leave.");
+        }
+
+        window.data()->setApp(d()->weak_from_this());
+
+        handle->push_back(window);
+    }
+
+    // The window is in the collection now; the widget waits here for the run
+    // loop to build its impl on the app thread.
```
samienne: Remove comment

src/bqui/src/app.cpp
```diff
@@ -457,75 +522,285 @@ class WindowGlue
...
+    return runUntil(std::move(running));
 }
 
-int App::run(bq::signal::AnySignal<bool> runningSignal) &&
+int App::run()
 {
-    ase::Platform platform = ase::makeDefaultPlatform();
+    // Run while a window is open, stopping when the last one leaves. A caller
+    // that wants to run past an empty collection passes its own signal to
+    // run(running).
```
samienne: Remove comment

src/bqui/src/app.cpp
```diff
@@ -457,75 +522,285 @@ class WindowGlue
...
+    auto mainQueue = context.getMainRenderQueue();
+
+    // The impls outlive this call — they are the app's, not the loop's — but
+    // the platform and render context they hold do not, so they must be
+    // released before this returns, however it returns.
+    ImplScope implScope { *d(), mainQueue };
+
+    // The running signal is the loop's only signal: the impls are a plain
+    // imperative collection synced below, not a signal, so it lives in its own
+    // context and is evaluated each frame.
```
samienne: Remove comment. Unnecessary details.

src/bqui/src/app.cpp
```diff
@@ -457,75 +522,285 @@ class WindowGlue
...
+    // queue addWindow fills.
+    auto sync = [&]()
     {
-        d()->windowGlues_.push_back(std::make_shared<WindowGlue>(
-            platform, context, std::move(w)));
-    }
+        std::vector<Window> windows = d()->getWindows();
+
+        // The next live set: surviving impls carried over, newly mounted ones
+        // added below. A window that survives keeps the impl it already had.
```
samienne: Remove comment

src/bqui/src/app.cpp
```diff
@@ -457,75 +522,285 @@ class WindowGlue
...
+        std::vector<btl::shared<WindowImpl>> next;
+        next.reserve(windows.size());
+
+        // Impls whose window has left. Held apart and destroyed last, after the
+        // live set is the new one.
```
samienne: Remove comment

src/bqui/src/app.cpp
```diff
@@ -457,75 +522,285 @@ class WindowGlue
...
+                next.push_back(std::move(impl));
+            else
+                departing.push_back(std::move(impl));
+        }
+
+        // Drain the pending mounts and build an impl for each window still in
+        // the collection that has none. A pending widget whose window has since
+        // left is dropped — a re-add re-supplies one — and one whose window is
+        // already mounted (a remove and re-add that has not yet cycled through
+        // a teardown) is dropped too.
```
samienne: Remove comment

src/bqui/src/app.cpp
```diff
@@ -457,75 +522,285 @@ class WindowGlue
...
+
+        // A frame that drew a departing window can still be in flight, and it
+        // holds that window's framebuffer, so nothing may release an impl until
+        // the queue has caught up.
+        if (!departing.empty())
+            mainQueue.finish();
+
+        // Swap the new complete set in, then clear the departed: destroying a
+        // window runs event handlers that reach back into the live set, which
+        // must be the new one by then.
```
samienne: Remove comment

src/bqui/src/app.cpp
```diff
@@ -457,75 +522,285 @@ class WindowGlue
...
+        d()->windowImpls_.swap(next);
+        departing.clear();
+    };
+
+    // Seed the initial windows before the first frame.
```
samienne: Remove comment

src/bqui/src/app.cpp
```diff
@@ -457,75 +522,285 @@ class WindowGlue
...
         {
             bq::signal::FrameInfo frame{ getNextFrameId(), aseFrame.dt };
-            running.update(frame);
 
-            return running.evaluate<0>().get<0>();
+            runningContext.update(frame);
+
+            // Sync the impls to the window collection after the running signal
+            // has updated, so a running signal derived from the collection
+            // observes it rather than intra-frame impl teardown.
```
samienne: Remove comment

src/bqui/src/windowdata.h
```diff
@@ -0,0 +1,103 @@
...
+            app_.reset();
+        }
+
+    private:
+        btl::UniqueId const id_ = btl::makeUniqueId();
+        bq::signal::AnySignal<std::string> title_;
+
+        // Guards the app reference and the callbacks alone. It is released
+        // before the app is touched, so close() never holds it while the app
+        // runs a window's destructor.
```
samienne: Remove comment

src/testapp1/main.cpp
```diff
@@ -69,6 +70,26 @@ std::vector<std::pair<std::string, avg::Curve>> curves = {
     { "easeInOutBounce", avg::curve::easeInOutBounce },
 };
 
+// The close button captures the owning Window and closes it. The window is a
+// small handle living in the app-owned widget, so this closes no cycle: the app
+// holds the widget, the widget holds the window, and the window points back at
+// the app only weakly.
```
samienne: Remove comment

src/testapp1/main.cpp
```diff
@@ -120,8 +141,58 @@ int main()
                 avg::RepeatMode::reverse
                 ));
 
+    // A tracked window the app owns, toggled by a button and closable by its
+    // own close button. 'showTracked' tracks whether it is open so the button
+    // reads the right label. The window is a persistent handle: its data and
+    // its onClose (which flips the flag) outlive each close, so a re-open
+    // supplies only a fresh widget. onClose is set once, here, not per-open, so
+    // it is not appended anew on every open.
```
samienne: Remove comment

src/testapp1/main.cpp
```diff
@@ -120,8 +141,58 @@ int main()
...
+                                w.close();
+                            })))
+                    | modifier::frame()
+                    | modifier::focusGroup());
+    };
+
     auto widgets = widget::hbox({
         widget::vbox({
+            // Every press opens another window. The app owns them, so each one
+            // closes by its own means and the app runs until none is left.
```
samienne: Remove comment

src/bqui/include/bqui/app.h
```diff
@@ -20,17 +25,74 @@ namespace bqui
...
+         */
+        int run(bq::signal::AnySignal<bool> running);
+
+        /** @overload
+         *
+         * Runs while the collection has a window in it, stopping when the last
+         * one is removed; an app with no window returns at once. This default
+         * policy is a 'running' signal derived from the collection — a caller
+         * that wants to outlive an empty collection passes its own signal to
+         * run(running).
```
samienne: No need to explain how it is implemented. It is enough to know that the app will stop once the last window is closed.

src/bqui/include/bqui/app.h
```diff
@@ -20,17 +25,74 @@ namespace bqui
...
+         * open windows, or finding one to remove.
+         */
+        std::vector<Window> getWindows() const;
+
+        /** @brief The app's windows, as a signal.
+         *
+         * The reactive form of the same collection, for a UI that follows it.
+         * It is an observation of the imperative collection, not its source:
+         * windows are still added and removed through addWindow()/
+         * removeWindow()/Window::close(), and this signal reports the result.
```
samienne: Signals are always just read only so no need to explain it here.

src/bqui/include/bqui/window.h
```diff
@@ -1,47 +1,82 @@
...
+     * callbacks): copies are one window and share that state, which lives as
+     * long as any copy does — across a remove and a later re-add. The window
+     * holds none of its contents; the widget is supplied to App::addWindow at
+     * mount and lives in the app-owned impl.
+     *
+     * So a widget inside the window may capture the very Window that owns it — a
+     * close button is @c [w]{ w.close(); } — with no retain cycle: the app holds
+     * the widget, the widget holds the Window, and the Window points back at the
+     * app only weakly.
+     */
```
samienne, line 24: It seems like this comment block is explaining a lot of the internal workings. The main idea of the library is that the user doesn't really have to care about the lifetimes or ownership. The user can do anything they want with the Window. Just briefly explain the usage, no reasoning or how it works.

PR #126, opened 2026-07-26
Add btl::RunLoop and render the platforms on demand

samienne:
Add `btl::RunLoop` and render the platforms on demand (the frame-model rework).

- `btl::RunLoop`: a single-threaded reactor (readable/writable/timer/post) with an opaque non-owning `NativeHandle`, split into a thread-safe owner + a loop-thread `Controller`. POSIX select + Win32 WSAEventSelect / message-queue / HANDLE sources.
- WGL/GLX/dummy drive their frame loop through it: block when idle, wake on `requestFrame()` (gated by the #127 observe mechanism), re-evaluate the signal graph only on external change; the desktop frame fence completes asynchronously instead of blocking.
- The dummy platform becomes a real headless backend: a time-advancing `Frame` plus an optional `setMaxFps()` cap.
- Also carries the runloop + frame-model design docs and the review-comment cleanups (explicit lambda captures, style fixes).

Green across the full CI matrix (Linux gcc/clang incl. Sanitize, macOS, Windows msvc + clang-cl). Base: #127.

### DIFF
--- a/src/ase/include/ase/dummyplatform.h
+++ b/src/ase/include/ase/dummyplatform.h
@@ -25,6 +25,7 @@ namespace ase
         // Set to true by requestFrame() (possibly off-thread) to coalesce a
         // burst of wake requests into a single posted task.
         std::atomic<bool> wakePosted_ = false;
+
         // Installed by run() so requestFrame(), which cannot see run()'s local
         // tick, can schedule one while the loop is active; cleared at run() exit.
         std::function<void(btl::RunLoop::Controller&)> scheduleTick_;
--- a/src/ase/include/ase/dummyplatform.h
+++ b/src/ase/include/ase/dummyplatform.h
@@ -4,6 +4,11 @@
 
 #include "asevisibility.h"
 
+#include <btl/runloop.h>
+
+#include <atomic>
+#include <functional>
+
 namespace ase
 {
     class ASE_EXPORT DummyPlatform : public PlatformImpl
@@ -15,6 +20,14 @@ namespace ase
         void run(RenderContext& renderContext,
                 std::function<bool(Frame const&)> frameCallback) override;
         void requestFrame() override;
+
+    private:
+        // Set to true by requestFrame() (possibly off-thread) to coalesce a
+        // burst of wake requests into a single posted task.
+        std::atomic<bool> wakePosted_ = false;
+        // Installed by run() so requestFrame(), which cannot see run()'s local
+        // tick, can schedule one while the loop is active; cleared at run() exit.
+        std::function<void(btl::RunLoop::Controller&)> scheduleTick_;
     };
 
 } // namespace ase
--- a/src/ase/include/ase/dummyplatform.h
+++ b/src/ase/include/ase/dummyplatform.h
@@ -21,6 +21,12 @@ namespace ase
                 std::function<bool(Frame const&)> frameCallback) override;
         void requestFrame() override;
 
+        /** @brief Cap the headless frame rate. Zero (the default) leaves the loop
+         * uncapped: a tick runs as soon as requestFrame() wakes it. A positive
+         * value paces ticks to at most fps frames per second. Set before run().
+         */
+        void setMaxFps(unsigned int fps);
+
     private:
         // Set to true by requestFrame() (possibly off-thread) to coalesce a
         // burst of wake requests into a single posted task.
@@ -29,6 +35,10 @@ namespace ase
         // Installed by run() so requestFrame(), which cannot see run()'s local
         // tick, can schedule one while the loop is active; cleared at run() exit.
         std::function<void(btl::RunLoop::Controller&)> scheduleTick_;
+
+        // Frame-rate cap; 0 means uncapped. Read only on the loop thread, so it
+        // must be set before run().
+        unsigned int maxFps_ = 0;
     };
 
 } // namespace ase
--- a/src/ase/include/ase/glxplatform.h
+++ b/src/ase/include/ase/glxplatform.h
@@ -4,11 +4,15 @@
 
 #include "asevisibility.h"
 
+#include <btl/runloop.h>
+
 #include <tracy/Tracy.hpp>
 
 #include <GL/glx.h>
 #include <X11/Xlib.h>
 
+#include <atomic>
+#include <functional>
 #include <string>
 #include <mutex>
 
@@ -65,6 +69,13 @@ namespace ase
         inline GlxPlatformDeferred* d() { return deferred_; }
         inline GlxPlatformDeferred const* d() const { return deferred_; }
         GlxPlatformDeferred* deferred_;
+
+        // Set to true by requestFrame() (possibly off-thread) to coalesce a
+        // burst of wake requests into a single posted task.
+        std::atomic<bool> wakePosted_ = false;
+        // Installed by run() so requestFrame(), which cannot see run()'s local
+        // tick, can schedule one while the loop is active; cleared at run() exit.
+        std::function<void(btl::RunLoop::Controller&)> scheduleTick_;
     };
 }
 
--- a/src/ase/include/ase/glxplatform.h
+++ b/src/ase/include/ase/glxplatform.h
@@ -73,6 +73,7 @@ namespace ase
         // Set to true by requestFrame() (possibly off-thread) to coalesce a
         // burst of wake requests into a single posted task.
         std::atomic<bool> wakePosted_ = false;
+
         // Installed by run() so requestFrame(), which cannot see run()'s local
         // tick, can schedule one while the loop is active; cleared at run() exit.
         std::function<void(btl::RunLoop::Controller&)> scheduleTick_;
--- a/src/ase/include/ase/wglplatform.h
+++ b/src/ase/include/ase/wglplatform.h
@@ -50,6 +50,7 @@ namespace ase
         // Set to true by requestFrame() (possibly off-thread) to coalesce a
         // burst of wake requests into a single posted task.
         std::atomic<bool> wakePosted_ = false;
+
         // Installed by run() so requestFrame(), which cannot see run()'s local
         // tick, can schedule one while the loop is active; cleared at run() exit.
         std::function<void(btl::RunLoop::Controller&)> scheduleTick_;
--- a/src/ase/include/ase/wglplatform.h
+++ b/src/ase/include/ase/wglplatform.h
@@ -4,10 +4,15 @@
 
 #include "asevisibility.h"
 
+#include <btl/runloop.h>
+
 #include <windows.h>
 
 #include <GL/gl.h>
 #include <GL/wglext.h>
+
+#include <atomic>
+#include <functional>
 
 namespace ase
 {
@@ -41,6 +46,13 @@ namespace ase
         HGLRC dummyContext_ = nullptr;
         HDC dummyDc_ = nullptr;
         PFNWGLCREATECONTEXTATTRIBSARBPROC wglCreateContextAttribsARB_ = nullptr;
+
+        // Set to true by requestFrame() (possibly off-thread) to coalesce a
+        // burst of wake requests into a single posted task.
+        std::atomic<bool> wakePosted_ = false;
+        // Installed by run() so requestFrame(), which cannot see run()'s local
+        // tick, can schedule one while the loop is active; cleared at run() exit.
+        std::function<void(btl::RunLoop::Controller&)> scheduleTick_;
     };
 
 } // namespace ase
--- a/src/ase/src/dummy/dummyplatform.cpp
+++ b/src/ase/src/dummy/dummyplatform.cpp
@@ -34,13 +34,11 @@ void DummyPlatform::run(RenderContext&,
 {
     Frame frame{};
 
-    // A tick is "scheduled" while one is posted; this loop-thread-only flag
-    // dedupes so ticks never stack.
     bool tickScheduled = false;
 
     std::function<void(btl::RunLoop::Controller&)> tick;
 
-    auto scheduleTick = [&](btl::RunLoop::Controller& controller)
+    auto scheduleTick = [&tickScheduled, &tick](btl::RunLoop::Controller& controller)
     {
         if (tickScheduled)
             return;
@@ -52,7 +50,7 @@ void DummyPlatform::run(RenderContext&,
     // render). It does not re-post; the loop blocks until requestFrame() wakes
     // it, so any sources registered on it (a remote socket, a timer) are
     // serviced meanwhile.
-    tick = [&](btl::RunLoop::Controller& controller)
+    tick = [&tickScheduled, &frameCallback, &frame](btl::RunLoop::Controller& controller)
     {
         tickScheduled = false;
 
@@ -63,9 +61,10 @@ void DummyPlatform::run(RenderContext&,
         }
     };
 
-    runLoop().post([&](btl::RunLoop::Controller& controller)
+    scheduleTick_ = [&scheduleTick](btl::RunLoop::Controller& c) { scheduleTick(c); };
+
+    runLoop().post([&scheduleTick](btl::RunLoop::Controller& controller)
         {
-            scheduleTick_ = [&](btl::RunLoop::Controller& c) { scheduleTick(c); };
             scheduleTick(controller);
         });
     runLoop().run();
--- a/src/ase/src/dummy/dummyplatform.cpp
+++ b/src/ase/src/dummy/dummyplatform.cpp
@@ -29,30 +29,64 @@ RenderContext DummyPlatform::makeRenderContext()
     return RenderContext(std::make_shared<DummyRenderContext>());
 }
 
+void DummyPlatform::setMaxFps(unsigned int fps)
+{
+    maxFps_ = fps;
+}
+
 void DummyPlatform::run(RenderContext&,
         std::function<bool(Frame const&)> frameCallback)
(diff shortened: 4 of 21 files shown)


src/btl/include/btl/nativehandle.h
```diff
@@ -0,0 +1,77 @@
...
+
+        alignas(alignof(std::max_align_t)) unsigned char
+            storage_[2 * sizeof(void*)] = {};
+        Kind kind_ = Kind::Invalid;
+    };
+
+    /** @brief The one door platform conversion code uses to fill or read a
+     * NativeHandle's bytes. Not for general use.
+     */
+    struct NativeHandleAccess
```
samienne: I sort of understand the urge to split these away from the NativeHandle but I still feel like having a class with static functions is a bit too much. I was thinking of making these still members of NativeHandle or free functions (although the friend machinery gets a bit trickier).

src/btl/src/posix/runloopimpl.cpp
```diff
@@ -0,0 +1,263 @@
...
+                std::function<void()> onWritable) override
+        {
+            return addFd(handle, std::move(onWritable), /*write=*/true);
+        }
+
+        TimerId addTimer(std::chrono::microseconds delay,
+                std::function<void()> callback) override
+        {
+            TimerId id = nextId_++;
+            timers_[id] = Timer{ Clock::now() + delay, std::move(callback) };
```
samienne, line 65: The run loop implementations must be fully thread safe as they can ve accessed from other threads as well. Optionally could be protected on platform as well.

src/btl/include/btl/runloop.h
```diff
@@ -0,0 +1,70 @@
...
+        ~RunLoop();
+
+        RunLoop(RunLoop&&) noexcept;
+        RunLoop& operator=(RunLoop&&) noexcept;
+
+        RunLoop(RunLoop const&) = delete;
+        RunLoop& operator=(RunLoop const&) = delete;
+
+        /** @brief Call onReadable while handle is readable. */
+        SourceId addReadable(NativeHandle handle, std::function<void()> onReadable);
```
samienne: Would be nice to have some kind of raii wrapper for these. And I was thinking that RunLoop would already return the wrapper, impls can operate on id level. The wrapper should have protection against deleted loop impl. E.g. Weak ptr. Maybe add a method to make the wrapper to drop the auto remove if the user doesn't want to store the wrapper to keep it alive. Add also the annotation to warn if the user doesn't use the return value.

src/btl/include/btl/runloop.h
```diff
@@ -0,0 +1,70 @@
...
+
+        /** @brief Call onWritable while handle is writable. */
+        SourceId addWritable(NativeHandle handle, std::function<void()> onWritable);
+
+        /** @brief Call callback once, after at least delay has elapsed. */
+        TimerId addTimer(std::chrono::microseconds delay,
+                std::function<void()> callback);
+
+        /** @brief Run task on the loop thread. Safe to call from any thread. */
+        void post(std::function<void()> task);
```
samienne: Actually just realized that this is the only thread safe method. I wonder if the whole api should be thread safe or just this function. I generally dislike having globally shared objects that are not thread safe so somekind of protection should be there. I wonder if the class should be split in two so that we get the non thread safe part as a reference to the post function callback. The impl could stay as is, both parts of the split would have the shared_ptr to the same impl.

src/btl/src/win32/runloopimpl.cpp
```diff
@@ -0,0 +1,286 @@
...
+        std::map<SourceId, Source> sources_;
+        std::map<TimerId, Timer> timers_;
+        std::mutex mutex_;
+        std::vector<std::function<void()>> posted_;
+        std::uint64_t nextId_ = 1;
+        bool running_ = false;
+    };
+}
+
+    std::unique_ptr<RunLoopImpl> makeRunLoopImpl()
```
samienne: These could be names something like makePlatformSpecificRunLoopImpl to highlight this is implemented for each platform separately.

src/btl/src/win32/runloopimpl.cpp
```diff
@@ -0,0 +1,286 @@
+// These must precede any header that pulls in winsock2/windows (below).
+#ifndef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN
+#endif
+#ifndef NOMINMAX
+#define NOMINMAX
+#endif
+
+#include "../runloopimpl.h"
```
samienne: Let's avoid .. in include paths. Just add src in the include directories.

src/ase/src/glx/glxplatform.cpp
```diff
@@ -413,7 +430,20 @@ void GlxPlatform::run(RenderContext& renderContext,
         }
 
         lastFrame = thisFrame;
-    }
+
+        controller.post(tick);
+    };
+
+    // Register the X source and start the frame loop on the loop thread; the
+    // Source is removed when xSource goes out of scope below.
```
samienne: Unnecessary comment

src/ase/src/glx/glxplatform.cpp
```diff
@@ -355,7 +358,18 @@ void GlxPlatform::run(RenderContext& renderContext,
     std::queue<btl::future::Future<>> frameFutures;
     auto mainQueue = renderContext.getMainRenderQueue();
 
-    while (true)
+    // Service the X connection through the run loop: when its socket has data,
+    // drain every queued event. handleEvents() is also called per frame below,
+    // which is what actually guarantees delivery today - a busy render loop, and
+    // Xlib can buffer events without new socket data (so select would not
+    // re-fire). This source prepares for on-demand pacing, where the loop blocks
+    // and the fd wakes it.
```
samienne: Unneeded comment

src/ase/src/windows/wglplatform.cpp
```diff
@@ -297,7 +298,13 @@ void WglPlatform::run(RenderContext& renderContext,
     std::queue<btl::future::Future<>> frameFutures;
     auto mainQueue = renderContext.getMainRenderQueue();
 
-    while (true)
+    // Drive frames through the run loop so sources registered on it (a future
+    // remote socket, timers) are serviced between frames. Each frame pumps the
+    // Win32 message queue, renders, then re-posts the next; a false callback
+    // stops the loop. The message pump stays here in ase - btl has no HWND
+    // awareness.
```
samienne: Remove comment.

src/ase/src/windows/wglplatform.cpp
samienne: Should we actually use .then and then do the post there? The thread now gets blocked and could do meaningful work. Although this might bring some latency when the wait actually happens.

src/ase/src/dummy/dummyplatform.cpp
```diff
@@ -33,8 +33,21 @@ void DummyPlatform::run(RenderContext&,
...
-    }
+
+    // Drive frames through the run loop so any sources registered on it (a
+    // remote socket, a timer) are serviced between frames. Each frame re-posts
+    // the next; a false return stops the loop.
+    std::function<void(btl::RunLoop::Controller&)> tick =
+        [&](btl::RunLoop::Controller& controller)
+    {
+        if (frameCallback(frame))
+            controller.post(tick);
```
samienne, line 68: Should the dummy platform be limited to some fps?

src/ase/include/ase/dummyplatform.h
```diff
@@ -15,6 +20,14 @@ namespace ase
         void run(RenderContext& renderContext,
                 std::function<bool(Frame const&)> frameCallback) override;
         void requestFrame() override;
+
+    private:
+        // Set to true by requestFrame() (possibly off-thread) to coalesce a
+        // burst of wake requests into a single posted task.
+        std::atomic<bool> wakePosted_ = false;
```
samienne, line 33: Put space after variable and before comments.

src/ase/include/ase/glxplatform.h
```diff
@@ -65,6 +69,13 @@ namespace ase
         inline GlxPlatformDeferred* d() { return deferred_; }
         inline GlxPlatformDeferred const* d() const { return deferred_; }
         GlxPlatformDeferred* deferred_;
+
+        // Set to true by requestFrame() (possibly off-thread) to coalesce a
+        // burst of wake requests into a single posted task.
+        std::atomic<bool> wakePosted_ = false;
```
samienne, line 75: Put space after this line

src/ase/include/ase/wglplatform.h
```diff
@@ -41,6 +46,13 @@ namespace ase
         HGLRC dummyContext_ = nullptr;
         HDC dummyDc_ = nullptr;
         PFNWGLCREATECONTEXTATTRIBSARBPROC wglCreateContextAttribsARB_ = nullptr;
+
+        // Set to true by requestFrame() (possibly off-thread) to coalesce a
+        // burst of wake requests into a single posted task.
+        std::atomic<bool> wakePosted_ = false;
```
samienne, line 52: Put space after this line.

src/ase/src/dummy/dummyplatform.cpp
```diff
@@ -33,12 +33,59 @@ void DummyPlatform::run(RenderContext&,
...
+
+    // Headless: a tick only runs the frame callback (there is nothing to
+    // render). It does not re-post; the loop blocks until requestFrame() wakes
+    // it, so any sources registered on it (a remote socket, a timer) are
+    // serviced meanwhile.
+    tick = [&](btl::RunLoop::Controller& controller)
+    {
+        tickScheduled = false;
+
+        if (!frameCallback(frame))
```
samienne, line 91: As the dummy platform is now evolving to be a real platform, I think we should inject a proper Frames instead of the same dummy one.

src/ase/src/dummy/dummyplatform.cpp
```diff
@@ -33,12 +33,59 @@ void DummyPlatform::run(RenderContext&,
...
+        {
+            controller.stop();
+            return;
+        }
+    };
+
+    runLoop().post([&](btl::RunLoop::Controller& controller)
+        {
+            scheduleTick_ = [&](btl::RunLoop::Controller& c) { scheduleTick(c); };
```
samienne: I'm not sure why this has to be done inside this lambda. The .run() runs the actual loop on line 71 so it should be ok to access the scheduleTick_ before calling run(). Make sure this is correct though.

src/ase/src/glx/glxplatform.cpp
```diff
@@ -352,82 +354,140 @@ void GlxPlatform::run(RenderContext& renderContext,
...
+                    });
 
-    while (!frameFutures.empty())
-    {
-        ZoneScopedN("Wait for frame to finish");
-        ZoneValue(frameFutures.size());
-        frameFutures.front().wait();
-        frameFutures.pop();
-    }
+            scheduleTick_ = [&](btl::RunLoop::Controller& c) { scheduleTick(c); };
```
samienne: This could be done outside of post.

src/bqui/src/windowimpl.h
```diff
@@ -75,6 +76,10 @@ class WindowImpl
         aseWindow.setVisible(true);
         aseWindow.setTitle(titleSignal_.evaluate<0>().get<0>());
 
+        auto wake = [this] { needsUpdate_ = true; aseWindow.requestFrame(); };
```
samienne: No one line lambdas

src/bqui/src/windowimpl.h
```diff
@@ -75,6 +76,10 @@ class WindowImpl
...
+        widgetInstanceSignal_.observe(wake);
+        titleSignal_.observe(wake);
+
         aseWindow.setFrameCallback([this](ase::Frame const& frame) {
```
samienne: Opening brace on its own line.

src/bqui/src/windowimpl.h
```diff
@@ -381,7 +393,11 @@ class WindowImpl
         auto timer = std::chrono::duration_cast<std::chrono::milliseconds>(
                 timer_);
 
-        makeTransaction(frame.dt, std::nullopt);
+        // Frames are scheduled to render and advance the render tree; the
+        // signal graph is re-evaluated only when the observe wake flagged an
+        // external change, not on every animation frame.
```
samienne: Remove unnecessary comment.

src/btl/include/btl/nativehandle.h
samienne: Already addressed on the branch: these are now free functions (makeNativeHandle / loadNativeHandle) befriended by NativeHandle, not a class of static functions.

src/btl/include/btl/runloop.h
samienne: Done: addReadable / addWritable / addTimer now return RAII Source / Timer handles (weak_ptr to the impl, remove/cancel on drop, detach() to opt out) and are marked [[nodiscard]].

src/btl/include/btl/runloop.h
samienne: Addressed: RunLoop now exposes only post() / stop() (thread-safe); the non-thread-safe registration API lives on Controller, handed to callbacks on the loop thread only. That is the two-part split - both sides share the impl via shared_ptr / weak_ptr.

src/btl/src/posix/runloopimpl.cpp
samienne: Addressed by that split: only post() / stop() cross threads, and post() locks mutex_ then wakes the loop via the self-pipe. Everything else runs on the loop thread through Controller, so the impl need not be fully thread-safe. See the contract comment in runloop.h.

src/btl/src/win32/runloopimpl.cpp
samienne: Already done - the factory is named makePlatformSpecificRunLoopImpl, matching the posix impl, the runloopimpl.h declaration and the runloop.cpp call site.

src/btl/src/win32/runloopimpl.cpp
samienne: Already done - btl/meson.build lists include_directories('include', 'src') and there are no '..' includes left under src/btl/src (win32 line 9 is a clean #include "runloopimpl.h").

src/ase/src/windows/wglplatform.cpp
samienne: Addressed - the fence path no longer blocks: pushFence() posts back to the loop to decrement framesInFlight, so the thread stays free while the GPU works rather than waiting.

src/ase/src/glx/glxplatform.cpp
samienne: Removed in f7a99b2 (dropped from all three platforms to keep them identical).

src/ase/src/glx/glxplatform.cpp
samienne: Removed in f7a99b2 (the re-arm comment is gone from wgl and glx).

src/ase/src/windows/wglplatform.cpp
samienne: Removed in f7a99b2 (this comment dropped from dummy/wgl/glx).

src/ase/include/ase/dummyplatform.h
samienne: Done in f7a99b2 - blank line after wakePosted_ in all three headers.

src/ase/include/ase/glxplatform.h
samienne: Done in f7a99b2.

src/ase/include/ase/wglplatform.h
samienne: Done in f7a99b2.

src/ase/src/dummy/dummyplatform.cpp
samienne: Done in f7a99b2 - the scheduleTick_ install is now set in run()'s body before the post; only the initial scheduleTick(controller) stays in the post since it needs the controller. Applied to all three platforms.

src/ase/src/glx/glxplatform.cpp
samienne: Done in f7a99b2 (moved out of the post on all three platforms, matching the dummy comment).

src/bqui/src/windowimpl.h
samienne: Done in f7a99b2 - expanded to a multi-line lambda body.

src/bqui/src/windowimpl.h
samienne: Done in f7a99b2 - opening brace on its own line (Allman).

src/bqui/src/windowimpl.h
samienne: Removed in f7a99b2.

src/ase/src/dummy/dummyplatform.cpp
samienne: Done in 01ed421 - added an optional setMaxFps() cap (0 = uncapped, the default). When capped, a tick due sooner than the interval is deferred by a one-shot timer; uncapped keeps the pure on-demand behaviour.

src/ase/src/dummy/dummyplatform.cpp
samienne: Done in 01ed421 - the dummy now builds a proper Frame{time, dt} from a steady_clock, mirroring the WGL/GLX platforms, instead of reusing an empty Frame{}.

PR #127, opened 2026-07-30
Fix signal::input observe and expose SignalContext::observe

samienne:
Rework signal observation - the groundwork for on-demand rendering (#126).

- Fix `signal::input` to be a real observe leaf: `InputHandle::set()` now fires the observe callbacks (was a no-op), mirroring `stream::collect`.
- Expose `SignalContext::observe(callback)`, a context-level external-change wakeup. The mechanism existed on the signal impls but had no public entry point.
- Hold observers by shared_ptr with eager deregistration; document the wakeup contract: always-armed, fires on external change, may fire more than once per cycle - a wakeup, not a change count.

No behaviour change for existing code (observe has no consumer on master yet). New `observetest.cpp` covers the leaves, forwarding through map/merge/combine/conditional/join/forEach, diamonds, and the context API. bq green on clang-cl.

### DIFF
--- a/src/bq/include/bq/signal/datacontext.h
+++ b/src/bq/include/bq/signal/datacontext.h
@@ -159,7 +159,7 @@ namespace bq::signal
         }
 
         /** The wakeup an external leaf registers itself with at init. */
-        std::weak_ptr<ObserveControl> observeControl() const
+        std::shared_ptr<ObserveControl> observeControl() const
         {
             return observeControl_;
         }
--- a/src/bq/include/bq/signal/datacontext.h
+++ b/src/bq/include/bq/signal/datacontext.h
@@ -8,11 +8,57 @@
 #include <typeindex>
 #include <memory>
 #include <unordered_map>
+#include <functional>
 #include <cassert>
 
 namespace bq::signal
 {
     BQ_EXPORT btl::UniqueId makeUniqueId();
+
+    /** Lifetime token for an observation. observe() ties every wakeup it
+     * registers to one of these; the wakeups fire while it is alive and drop
+     * when it does. */
+    using ObserveGuard = std::shared_ptr<void>;
+
+    /** Creates a fresh guard for a caller about to observe. */
+    inline ObserveGuard makeObserveGuard()
+    {
+        return std::make_shared<char>();
+    }
+
+    /** A wake callback paired with the guard that scopes it.
+     *
+     * observe() stores these on the external leaves it reaches rather than
+     * returning an unregistration handle: the callback fires only while its
+     * guard is alive, so an observer disarms by dropping the guard. A leaf drops
+     * a dead registration the next time it is observed or fires. */
+    class ObserveCallback
+    {
+    public:
+        ObserveCallback() = default;
+
+        ObserveCallback(ObserveGuard const& guard, std::function<void()> callback) :
+            guard_(guard),
+            callback_(std::move(callback))
+        {
+        }
+
+        /** True while the owning observer is still armed. */
+        explicit operator bool() const
+        {
+            return !guard_.expired();
+        }
+
+        void operator()() const
+        {
+            if (auto alive = guard_.lock())
+                callback_();
+        }
+
+    private:
+        std::weak_ptr<void> guard_;
+        std::function<void()> callback_;
+    };
 
     class BQ_EXPORT DataContext
     {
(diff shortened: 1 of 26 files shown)


src/bq/include/bq/signal/signalcontext.h
```diff
@@ -115,6 +136,14 @@ namespace bq::signal
             return result;
         }
 
+        // The callback is copied into each leaf.
```
samienne: Remove comment

src/bq/include/bq/signal/input.h
```diff
@@ -60,30 +66,62 @@ namespace bq::signal
...
-                control->sigData = control->sig->initialize(control->context);
-                control->value = control->sig->evaluate(control->context,
-                        *control->sigData);
-                ++control->index;
-                */
+                for (auto it = control->observers.begin();
+                        it != control->observers.end(); )
+                {
+                    if (it->second.expired())
+                        it = control->observers.erase(it);
```
samienne: The callbacks are guarded by the guards in the datacontext so these should already be cleaned when the data context is cleaned. This also happens when the signal graph structure changes in a way the signal is no longer part of the context.

src/bq/include/bq/signal/datacontext.h
```diff
@@ -8,12 +8,41 @@
...
+        void arm(std::function<void()> callback)
+        {
+            std::lock_guard<std::mutex> lock(mutex_);
+            callback_ = std::move(callback);
+        }
+
+        void fire()
+        {
+            std::function<void()> callback;
+            { std::lock_guard<std::mutex> lock(mutex_); callback = callback_; }
```
samienne: Split to multiple lines.

src/bq/include/bq/signal/input.h
```diff
@@ -60,30 +66,62 @@ namespace bq::signal
...
-                std::unique_lock lock(control->mutex);
+                std::unique_lock<std::recursive_mutex> lock(control->mutex);
                 control->value = SignalResult<Ts...>(std::forward<Us>(us)...);
                 control->valueIndex = std::max(control->valueIndex,
                         control->signalIndex) + 1;
+                for (auto it = control->observers.begin();
+                        it != control->observers.end(); )
+                {
+                    if (it->second.expired())
+                        it = control->observers.erase(it);
```
samienne: These should be cleared by the guard that is in the datacontext.

src/bq/include/bq/signal/datacontext.h
```diff
@@ -118,11 +147,26 @@ namespace bq::signal
             frameData_.clear();
         }
 
+        /** Arms this context's wakeup; the callback fires on the next external
+         * change to any leaf reached by this context. */
+        void observe(std::function<void()> callback)
```
samienne: I would name this as something like setObserveCallback or similar to highlight that it overrides the previously set callback. I don't think we need to support multiple callbacks since the only use case is that signal context has the single callback registered.

src/bq/include/bq/signal/signalcontext.h
```diff
@@ -71,6 +71,14 @@ namespace bq::signal
             return false;
         }
 
+        /** Arms a wakeup for external changes to any leaf this context reaches.
+         * The callback fires without the graph being evaluated. It stays armed
+         * for the context's lifetime. */
+        void observe(std::function<void()> callback)
```
samienne, line 86: I would note that it is possible to get multiple calls per update (even for single set if the signal graph has some kind of a diamond pattern).

src/bq/include/bq/signal/input.h
```diff
@@ -60,30 +74,62 @@ namespace bq::signal
...
+                for (auto it = control->observers.begin();
+                        it != control->observers.end(); )
+                {
+                    if (auto registration = it->lock())
+                    {
+                        toFire.push_back(std::move(registration));
+                        ++it;
+                    }
+                    else
+                        it = control->observers.erase(it);
```
samienne: Again, I think these should be already cleared by the DataContext guards. The weak_ptr should never go stale (I'm even wondering if we should go with a shared_ptr instead to avoid the lock()).

src/bq/include/bq/signal/input.h
```diff
@@ -60,30 +74,62 @@ namespace bq::signal
...
```
samienne: Here the same, the guard should already handle these.

